PR #7, opened 2019-12-12
Update button and icon size

stredger:
## Description
Had to do some weirdness with height and margins as increasing the icon size made the buttons sit weird... (see screenshots) if you know a better way to do this or have seen something similar before hit me up

## Screenshots
before
Screen Shot 2019-12-12 at 9 38 28 AM

after
Screen Shot 2019-12-12 at 9 40 06 AM

### DIFF
--- a/packages/riipen-ui/src/components/Button.jsx
+++ b/packages/riipen-ui/src/components/Button.jsx
@@ -179,10 +179,10 @@ class Button extends React.Component {
           }
 
           .icon-start {
-            margin-right: 10px;
+            margin-right: ${theme.spacing(2)}px;
           }
           .icon-end {
-            margin-left: 10px;
+            margin-left: ${theme.spacing(2)}px;
           }
 
           .full-width {
--- a/packages/riipen-ui/src/components/Button.jsx
+++ b/packages/riipen-ui/src/components/Button.jsx
@@ -131,13 +131,13 @@ class Button extends React.Component {
         >
           <span className={clsx("label")}>
             {IconStart && (
-              <span className={clsx("icon-start")}>
+              <span className={clsx("icon", "icon-start")}>
                 <IconStart />
               </span>
             )}
             {children}
             {IconEnd && (
-              <span className={clsx("icon-end")}>
+              <span className={clsx("icon", "icon-end")}>
                 <IconEnd />
               </span>
             )}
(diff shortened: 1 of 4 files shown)


packages/riipen-ui/src/components/Button.jsx
```diff
@@ -165,19 +165,24 @@ class Button extends React.Component {
...
           a.root {
-            min-width: 70px;
+            min-width: 100px;
             text-decoration: none;
           }
 
+          .icon {
+            font-size: 16px;
+            height: 13px;
+            margin-top: -3px;
```
jordanell: Are you sure this is right? Do font awesome icons have white space at the top of them? Super frustrating if true.

packages/riipen-ui/src/components/Button.jsx
```diff
@@ -165,19 +165,24 @@ class Button extends React.Component {
...
+          }
+          .icon > svg {
+            display: flex;
+          }
+
           .icon-start {
-            margin-right: 8px;
+            margin-right: 10px;
```
jordanell: I wonder if this should be `${theme.spacing(2)}` instead?

packages/riipen-ui/src/components/Button.jsx
```diff
@@ -165,19 +165,24 @@ class Button extends React.Component {
...
           }
           .icon-end {
-            margin-left: 8px;
-          }
-          .icon-start > svg,
-          .icon-end > svg {
-            display: flex;
+            margin-left: 10px;
```
jordanell: Same here.

packages/riipen-ui/src/components/Button.jsx
stredger: They dont, but im trying to keep the icons the same size as the text font size. If the icons are bigger then the buttons grow in different ways ie. if we have an icon before the text the button grows upwards, while if we have an icon after the text the button grows down. this is the only way i could make them line up (tried a bunch of flex stuff but nothing seemed to help). Maybe we just live with the buttons being slightly different and handle it in the container?

stredger: given that whatever is in the button can change size as well

jordanell: Ah ok that makes sense. Nah I think this is fine rather than trying anything else.

PR #8, opened 2019-12-12
Fix radio buttons on border-box site

kradical:
## Description
All radio buttons on our site look like 💩 right now because our site is border-box but the ui demo site is content-box.
## Notes
huge change
## Screenshots
before:
Screen Shot 2019-12-12 at 2 24 46 PM

after:
Screen Shot 2019-12-12 at 2 24 10 PM


kradical:
I realized I didn't bump the version, I'll just let the next version bump take this in.

PR #12, opened 2020-01-27
Popover fix

bolinkd:
## Description
Updated the popover for the menu to be rendered as hidden in the top left corner until it is opened. 
This fixes the issue where the popover would not be in the correct position if it was initially positioned too far to the right of the screen

## Notes
Turns out position: absolute combined with top and left attributes is 'smart' enough to make sure the width of the box is not more than the width of the view window. So I have positioned the box in the top left to start and then move it to the correct location when the popover is opened.

### DIFF
--- a/packages/riipen-ui/src/components/MenuList.jsx
+++ b/packages/riipen-ui/src/components/MenuList.jsx
@@ -155,18 +155,11 @@ class MenuList extends React.Component {
   render() {
     const { children, classes, selectedIndex } = this.props;
 
-    const theme = this.context;
-
     return (
       <React.Fragment>
         <div className={clsx(classes)}>
           {this.getListItems(children, selectedIndex)}
         </div>
-        <style jsx>{`
-          div {
-            margin: ${theme.spacing(1)}px 0;
-          }
-        `}</style>
       </React.Fragment>
     );
   }
--- a/packages/riipen-ui/src/components/Popover.jsx
+++ b/packages/riipen-ui/src/components/Popover.jsx
@@ -92,12 +92,19 @@ class Popover extends React.Component {
     this.state = {
       scrollContainer: null,
       scrollContainerStyle: null,
-      contentRef: null
+      contentStyles: {
+        top: 0,
+        left: 0,
+        visibility: "hidden"
+      }
     };
+
+    this.contentRef = React.createRef();
   }
 
   componentDidMount() {
     const { anchorEl, lockScroll } = this.props;
+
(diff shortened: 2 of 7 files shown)


packages/riipen-ui/src/components/MenuList.jsx
```diff
@@ -155,18 +155,11 @@ class MenuList extends React.Component {
...
     return (
       <React.Fragment>
         <div className={clsx(classes)}>
           {this.getListItems(children, selectedIndex)}
         </div>
-        <style jsx>{`
-          div {
-            margin: ${theme.spacing(1)}px 0;
-          }
-        `}</style>
```
kradical: I'm guessing this fixes a spacing issue that didn't exist before?

kradical: I'm guessing this fixes a spacing issue that didn't exist before?

packages/riipen-ui/src/components/Popover.jsx
```diff
@@ -305,11 +331,13 @@ class Popover extends React.Component {
...
             outline: 0;
-            overflow-x: hidden;
-            overflow-y: auto;
             position: absolute;
             z-index: ${theme.zIndex.middle};
           }
+
+          .popover::webkit-scrollbar {
+            display: none;
+          }
```
kradical: ahh the scrollbar fix? :D

packages/riipen-ui/src/components/Popover.jsx
```diff
@@ -126,7 +156,7 @@ class Popover extends React.Component {
   };
 
   getPositioningStyle = () => {
```
kradical: Maybe rename this method to `setPosition` or `updatePosition` or `setPositionStyle` etc.etc. something with set or update now that it is setting / updating.

packages/riipen-ui/src/components/MenuList.jsx
bolinkd: Screen Shot 2020-01-27 at 10 47 03 AM
It fixes this tiny annoyance :)

packages/riipen-ui/src/components/Popover.jsx
bolinkd: 👍

packages/riipen-ui/src/components/Popover.jsx
bolinkd: The main fix is in the overflow css changes above, this is just a final check to make sure as the scrollbar width is not included when the calculations are run. Long story short.... this **should** never happen xD

PR #13, opened 2020-02-18
[PLT-4660] Chip Component

bolinkd:
## Description
Added the chip component to the UI Kit that follows the old web pill component layout

## Screenshots
Screen Shot 2020-02-18 at 15 09 36
Screen Shot 2020-02-18 at 15 09 45

## Where to Start
packages/riipen-ui/src/components/Chip.jsx


packages/riipen-ui/src/components/Chip.jsx
```diff
@@ -68,103 +99,239 @@ class Chip extends React.Component {
...
             display: inline-flex;
             font-family: ${theme.typography.fontFamily};
             font-size: 13px;
-            height: 32px;
             justify-content: center;
+            ${spacing !== 0 && `margin-bottom: ${theme.spacing(spacing)}px;`}
+            ${spacing !== 0 &&
+              `margin-right: ${theme.spacing(
+                spacing
+              )}px;`}
```
bolinkd: This allows the user to overwrite the margins by providing spacing = 0

packages/riipen-ui/src/components/Chip.jsx
```diff
@@ -2,13 +2,15 @@ import clsx from "clsx";
...
 
-import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import { faTimes } from "@fortawesome/free-solid-svg-icons";
-
 import ThemeContext from "../styles/ThemeContext";
 
 class Chip extends React.Component {
   static propTypes = {
+    /**
+     * The elements to render inside the pill
```
jordanell: Small nitpick, but add a period to the end of this (and all subsequent prop comments) since they are rendered in the docs as full sentences.

packages/riipen-ui/src/components/Chip.jsx
jordanell: All "base" UI elements shouldn't have margins (in most cases) by default. It should be up to the application to determine how to render the element if it needs external spacing. I think we should remove this margin code. It also would eliminate the need for the `spacing` prop which doesn't really fit the types of props we want in the UI kit.

packages/riipen-ui/src/components/Chip.jsx
```diff
@@ -30,21 +40,41 @@ class Chip extends React.Component {
...
 
+    /**
+     * Whether to invert colors on hover
+     */
+    hover: PropTypes.bool,
+
+    /**
+     * Icon to display at start of chip
+     */
+    iconStart: PropTypes.elementType,
```
jordanell: Since you can't have an "end" icon, I would be tempted to rename this to just "icon".

packages/riipen-ui/src/components/Chip.jsx
```diff
@@ -30,21 +40,41 @@ class Chip extends React.Component {
...
-     * If set, the delete icon will be shown.
+     * Action to perform when clicked
+     */
+    onClick: PropTypes.func,
+
+    /**
+     * Callback function fired when the icon is clicked.
      */
-    onDelete: PropTypes.func,
+    onIconClick: PropTypes.func,
```
jordanell: Since this is just named `onIconClick` it makes sense to also rename the `iconStart` to just `icon`.

packages/riipen-ui/src/components/Chip.jsx
```diff
@@ -30,21 +40,41 @@ class Chip extends React.Component {
      */
     disabled: PropTypes.bool,
 
+    /**
+     * Whether to invert colors on hover
+     */
+    hover: PropTypes.bool,
```
jordanell: This props is slightly confusing to me. If I mouse over an element and it changes color, my brain tells me it should be clickable or interactable (apparently that isn't a real world) in some way. This doesn't seem to be the case here though. It looks like from the code it's just an aesthetic thing? This seems weird to me, but it's probably more of a Maribel question at this point. Can you verify that my thought process here is correct?

bolinkd: removed as it wasnt even being used.

PR #14, opened 2020-02-20
UI Kit Table

bolinkd:
## Description
This introduces two new functions to the UI Kit,
The first is the ability to create tables using the components of the library with easier to use props
The second (and far superior function) is too add a fancy automatic mobile friendly table!!!

The table renders itself using the breakpoints and columns provided (rather than rows). This allows it to break data and render it in a nice mobile view

## Notes
This is a first go, naming is not super important yet. The main thing is usability and readability. 

## Screenshots
Simple table:
Screen Shot 2020-02-20 at 15 54 27

Mobile Table Large:
Screen Shot 2020-02-20 at 15 55 55

Mobile Table Condensed: 
Screen Shot 2020-02-20 at 15 55 21

## Where to Start
If you want to see implementation: 
 - packages/riipen-ui/src/components/Table.jsx
If you want to see examples of usage:
 - packages/riipen-ui-docs/src/pages/components/table/Simple.jsx
 - packages/riipen-ui-docs/src/pages/components/table/Mobile.jsx


packages/riipen-ui-docs/src/pages/components/table/Mobile.jsx
```diff
@@ -0,0 +1,53 @@
+import React from "react";
+
+import Table from "@riipen-ui/components/Table";
+import TableColumn from "@riipen-ui/components/TableColumn";
+
+export default function Mobile() {
```
jordanell: Is there anyway we can simplify this documentation? I've read it 3 times and still feel like I don't totally know what's going on. Documentation should be verbose over clever to show what's really going on IMO.

packages/riipen-ui/src/components/Table.jsx
```diff
@@ -0,0 +1,241 @@
...
+import TableBody from "./TableBody";
+import TableCell from "./TableCell";
+import TableHeaderCell from "./TableHeaderCell";
+
+class Table extends React.Component {
+  static propTypes = {
+    /**
+     * Displays the label instead of the header
+     */
+    allowSearch: PropTypes.bool,
```
jordanell: This is very confusing to me. We have a prop that is called `allowSearch` but it's dealing with labels? Can we rename or update the docs to be slightly more clear as to what is happening?

packages/riipen-ui/src/components/Table.jsx
```diff
@@ -0,0 +1,241 @@
...
+
+    /**
+     * Classes to apply to the root element
+     */
+    classes: PropTypes.arrayOf(PropTypes.string),
+
+    /**
+     * The entities to render in the table
+     */
+    entities: PropTypes.array,
```
jordanell: What are table "entities"? Can we make this documentation more clear. Are these rows?

packages/riipen-ui/src/components/Table.jsx
```diff
@@ -0,0 +1,241 @@
...
+
+    /**
+     * Whether or not to highlight rows on highlight
+     */
+    hover: PropTypes.bool,
+
+    /**
+     * Size to change table render from desktop to mobile
+     */
+    mobileBreakpoint: PropTypes.oneOf(["xs", "sm", "md", "lg", "xl"])
```
jordanell: Nice!

packages/riipen-ui/src/components/TableCell.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Classes to apply to the td element
+     */
+    classes: PropTypes.arrayOf(PropTypes.string),
+
+    /**
+     * Padding to apply to the td element
+     */
+    spacing: PropTypes.number,
```
jordanell: IMO this should be removed. All our tables should have the same spacing.

If we keep this, I think we should rename to `padding`. The UI kit's props should be as close to the CSS attribute they represent as possible when it comes to naming.

packages/riipen-ui/src/components/TableCell.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Alignment property to apply to the td element
+     */
+    align: PropTypes.oneOf(["center", "left", "right", "justify", "char"]),
```
jordanell: Nice.

packages/riipen-ui/src/components/TableColumn.jsx
```diff
@@ -0,0 +1,31 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+class TableColumn extends React.Component {
```
jordanell: I am confused. Does this component not render anything? Didn't even think this was possible.

packages/riipen-ui/src/components/TableColumn.jsx
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * A callback function that returns
+     * the header cell to render for the column
+     */
+    header: PropTypes.func,
+
+    /**
+     * The label to display when not rendering the header
+     */
+    label: PropTypes.string
```
jordanell: Is this required?

packages/riipen-ui/src/components/TableContainer.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Whether to render the background for a table
+     */
+    background: PropTypes.bool,
+
+    /**
+     * The padding to add to the table element
+     */
+    spacing: PropTypes.number
```
jordanell: IMO remove again, but if not rename to `padding`.

packages/riipen-ui/src/components/TableContainer.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Classes to apply to the root element
+     */
+    classes: PropTypes.arrayOf(PropTypes.string),
+
+    /**
+     * Whether to give all columns in the table equal width
+     */
+    fixed: PropTypes.bool,
```
jordanell: I think this should be renamed to `tableLayout` and should have the whitelisted options available in the CSS.

packages/riipen-ui/src/components/TableContainer.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Whether to center the table in the parent container
+     */
+    centered: PropTypes.bool,
```
jordanell: I think this should be removed. It should be up to the app to determine if the table is centered in its application.

packages/riipen-ui/src/components/TableContainer.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Whether to render the background for a table
+     */
+    background: PropTypes.bool,
```
jordanell: Rename to `backgroundColor` and white list the available options (ie, "grey100", "transparent") and make the default prop value "grey100"

packages/riipen-ui/src/components/TableRow.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Classes to apply to the root element
+     */
+    classes: PropTypes.arrayOf(PropTypes.string),
+
+    /**
+     * The number of columns the row should span
+     */
+    span: PropTypes.number,
```
jordanell: Rename to `rowSpan`.

packages/riipen-ui/src/components/TableRow.jsx
```diff
@@ -0,0 +1,78 @@
...
+            ${border
+              ? `background: linear-gradient(
+                to bottom,
+                transparent calc(100% - 1px),
+                ${theme.palette.grey[400]} 0%,
+                ${theme.palette.grey[400]}
+              )
+              bottom / calc(100% - ${theme.spacing(8)}px);`
+              : ""}
+            border-left: 3px solid transparent;
```
jordanell: I am confused. What is this doing exactly?

packages/riipen-ui/src/components/TableColumn.jsx
```diff
@@ -0,0 +1,31 @@
...
+  };
+
+  static defaultProps = {
+    cell: () => {},
+    header: () => {}
+  };
+}
+
+export default TableColumn;
```
kradical: okok back to my comment about how we pass in column configuration:

I think this is a pink flag, a react component that doesn't do anything. We are declaring an interface: "hey this is how we will pass in table column info".  But then that component is torn apart and never rendered. I'll think on this one overnight because I do like the way passing in JSX columns looks, but it seems a bit.. not quite right.

packages/riipen-ui/src/components/TableHeader.jsx
```diff
@@ -0,0 +1,35 @@
...
+     */
+    children: PropTypes.node,
+
+    /**
+     * Classes to apply to the root element
+     */
+    classes: PropTypes.arrayOf(PropTypes.string)
+  };
+
+  static defaultProps = {};
```
kradical: We can remove this empty default props!

packages/riipen-ui/src/components/TableHeader.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+  static contextType = ThemeContext;
+
+  render() {
+    const { children, classes } = this.props;
+    return (
+      <React.Fragment>
+        <thead className={clsx(classes)}>{children}</thead>
+        <style jsx>{``}</style>
```
kradical: same comment here: Is this necessary? if not 👋

packages/riipen-ui/src/components/TableHeaderCell.jsx
```diff
@@ -0,0 +1,33 @@
...
+  };
+
+  static contextType = ThemeContext;
+
+  render() {
+    const { children, classes } = this.props;
+    return (
+      <React.Fragment>
+        <th className={clsx(classes)}>{children}</th>
+        <style jsx>{``}</style>
```
kradical: ❓ ➡️  👋

packages/riipen-ui/src/components/TableHeaderCell.jsx
```diff
@@ -0,0 +1,33 @@
...
+     */
+    children: PropTypes.node,
+
+    /**
+     * Classes to apply to the root element
+     */
+    classes: PropTypes.arrayOf(PropTypes.string)
+  };
+
+  static contextType = ThemeContext;
```
kradical: unused context!

packages/riipen-ui/src/components/TableHeader.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Classes to apply to the root element
+     */
+    classes: PropTypes.arrayOf(PropTypes.string)
+  };
+
+  static defaultProps = {};
+
+  static contextType = ThemeContext;
```
kradical: unused context!

packages/riipen-ui/src/components/TableRow.jsx
kradical: same here, possibly rename to `rowSpan`?

packages/riipen-ui/src/components/TableRow.jsx
bolinkd: This is how you have to render lines between rows in css. you can't apply border-bottom to a tr element, so instead you have to create a background that looks like a border. This one is taken directly from the Table component in the web repo

kradical: weirdness, it seems like if you add `border-collapse: collapse` to the table you can directly apply borders to `tr` elements. I tested it out in a sandbox and seems to work. Might be worth testing out to clean this up?

packages/riipen-ui/src/components/Table.jsx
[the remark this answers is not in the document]
bolinkd: so, one thing I still need to add in.

If you are on mobile and this is true, render the headers above the table. if false, there would be no headers rendered.

we could make a separate option for each, the problem with that is that if you have search criteria on a desktop and then move over to a mobile sized screen the search criteria would no longer be available to be edited (this is something I ran into when porting the web tables to use the UI Kit)

bolinkd: yeah, maybe something like useLabels or renderHeader. We could also say that if you want to disable search functionality it is the job of the website to pass in the component. (Similar to the way the members table does it now) I just thought it would be nice to have it in the UI kit with a simple toggle

packages/riipen-ui/src/components/Table.jsx
```diff
@@ -0,0 +1,241 @@
...
+     */
+    allowSearch: PropTypes.bool,
+
+    /**
+     * The children columns to render
+     * Must be of type TableColumn
+     */
+    children: PropTypes.oneOfType([
+      PropTypes.arrayOf(
+        PropTypes.shape({ type: PropTypes.oneOf([TableColumn]) })
```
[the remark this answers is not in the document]
bolinkd: Yeah, this was just from building out the different ways you could pass in column information. This would change to instanceOf

packages/riipen-ui/src/components/Table.jsx
bolinkd: we could do something like rowData or rowEnitities, this is the array of data that gets looped through to generate the table. Each value in the array is used in the call back for the cell render function

packages/riipen-ui/src/components/Table.jsx
```diff
@@ -0,0 +1,241 @@
...
+    allowSearch: true
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      isMobile: false
+    };
+
+    this.updateWindowDimensions = this.updateWindowDimensions.bind(this);
```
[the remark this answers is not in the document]
bolinkd: Ill double check, I know I have run into this in the past where for some reason things dont want to bind correctly and it fails to remove the event listener. Will also add a debounce at some point to make it more efficient

packages/riipen-ui/src/components/Table.jsx
```diff
@@ -0,0 +1,241 @@
...
+        Object.assign(acc, {
+          [child.key]: {
+            label: child.props.label,
+            primary: child.props.primary,
+            header: child.props.header,
+            cell: child.props.cell
+          }
+        }),
+      {}
+    );
```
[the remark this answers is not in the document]
bolinkd: so the reason I used key was because it was something that would be required in this case as you are creating a list of nodes in the parent component. I am fine with creating our own, but it would kind of be like adding a second key prop that would be used just for us and would 99% of the time be the exact same as the key value. 

This is the area I expected the most push back. I actually really like the way this is implemented as from the outside it is very easy to read. We could move all of this into a prop of the table component and just have one giant prop list but to me this feels better.

packages/riipen-ui/src/components/TableBody.jsx
```diff
@@ -0,0 +1,33 @@
...
+     */
+    children: PropTypes.node,
+
+    /**
+     * Classes to apply to the root element
+     */
+    classes: PropTypes.arrayOf(PropTypes.string)
+  };
+
+  static contextType = ThemeContext;
```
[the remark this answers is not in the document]
bolinkd: most of these are basic copy paste, will need some clean up but I am currently more worried about the implementation of the generated table than the clean up in these components. I will be sure to get to these when I clean it up

packages/riipen-ui/src/components/TableCell.jsx
bolinkd: noted

packages/riipen-ui/src/components/TableColumn.jsx
bolinkd: Yeah, You can render nothing by returning null which is the default when not implemented. I am not opposed to the idea of using a node to store information used by other components to render something as that is essentially what a HOC is. But that is my opinion on this, Id love to get other people feedback. 
This allows us to pass the rendering data in the children of the Table component rather than as an additional prop. Both are valid but I prefer this way.

packages/riipen-ui/src/components/TableColumn.jsx
bolinkd: I am not opposed to the idea of using a node to store information used by other components to render something as that is essentially what a HOC is. But that is my opinion on this, Id love to get other people feedback. 
This allows us to pass the rendering data in the children of the Table component rather than as an additional prop. Both are valid but I prefer this way.

packages/riipen-ui/src/components/TableRow.jsx
bolinkd: thanks, ill take a look

packages/riipen-ui/src/components/Table.jsx
jordanell: I think search should actually be up to the using application since it may work differently for different sites.

packages/riipen-ui/src/components/Table.jsx
jordanell: I like `rowData`

packages/riipen-ui/src/components/TableColumn.jsx
jordanell: I think this is fine.

packages/riipen-ui/src/components/Table.jsx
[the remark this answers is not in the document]
stredger: might want to test this with a react fragment as well. it might not like it

packages/riipen-ui/src/components/Table.jsx
kradical: I think even just "data", I was looking at other table implementations and react-tables takes in "data" and "columns" to construct everything you need.

packages/riipen-ui/src/components/Table.jsx
```diff
@@ -0,0 +1,241 @@
...
+    const { entities, children, classes } = this.props;
+
+    const theme = this.context;
+
+    let nodes = children;
+    if (!children.length) {
+      nodes = [children];
+    }
+
+    const tableContent = nodes.reduce(
```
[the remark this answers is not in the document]
bolinkd: 🙌

packages/riipen-ui/src/components/TableBody.jsx
```diff
@@ -0,0 +1,28 @@
...
+     */
+    classes: PropTypes.arrayOf(PropTypes.string)
+  };
+
+  render() {
+    const { children, classes } = this.props;
+    return (
+      <React.Fragment>
+        <tbody classes={clsx(classes)}>{children}</tbody>
+      </React.Fragment>
```
kradical: we can get rid of this fragment!

kradical: we can get rid of this fragment 👋

packages/riipen-ui/src/components/TableGenerator.jsx
```diff
@@ -0,0 +1,277 @@
...
+
+    const mobileHeaderIndex = bodyColumns.findIndex(x => x.mobileHeader);
+    if (mobileHeaderIndex !== -1) {
+      mobileHeader = bodyColumns.splice(mobileHeaderIndex, 1)[0];
+    }
+
+    const mobileFooterIndex = bodyColumns.findIndex(x => x.mobileFooter);
+    if (mobileFooterIndex !== -1) {
+      mobileFooter = bodyColumns.splice(mobileFooterIndex, 1)[0];
+    }
```
kradical: personal preference: I don't like splice. it mutates and hardly ever gets used (I had to look at the mdn docs to see what the second param was, and what it returns!). In general I find it makes code harder to follow then it otherwise would be.

If I got the logic right you can do something like:

```javascript
const mobileHeader = columns.find(x => x.mobileHeader);
const mobileFooter = columns.find(x => x.mobileFooter);
const bodyColumns = columns.filter(x => !x.mobileFooter && !x.mobileHeader);
```

Which IMO is a bit more clear.

packages/riipen-ui/src/components/TableHeader.jsx
```diff
@@ -0,0 +1,28 @@
...
+     */
+    classes: PropTypes.arrayOf(PropTypes.string)
+  };
+
+  render() {
+    const { children, classes } = this.props;
+    return (
+      <React.Fragment>
+        <thead className={clsx(classes)}>{children}</thead>
+      </React.Fragment>
```
kradical: unnecessary Fragment 👋

packages/riipen-ui/src/components/TableGenerator.jsx
bolinkd: you see, I was thinking the same thing... and for some reason [].filter just slipped my mind.. It was late lol

packages/riipen-ui/src/components/TableHeader.jsx
bolinkd: I didnt like him anyways

packages/riipen-ui/src/components/TableBody.jsx
bolinkd: who does this guy think he is... 👋

PR #15, opened 2020-02-26
Add withClasses HOC to all components

bolinkd:
## Description
The new plan is to include a "riipen" class with all components exposed in the UIKit. 
In order to override styles on the web a class selector like the one below can be used.
:global(.riipen).myNewClass {}
This allows us to override defaults while still making changes to the UI Kit obvious.
The general rule for the UI Kit should be:
” oh, you used :global(.riipen), why was that needed? should something be added to the kit? is it a one off?”

## Notes
I have used a HOC to allow us to add more custom classes using the component name in the future if we deem it necessary.

## Screenshots
Screen Shot 2020-02-26 at 15 32 54
Screen Shot 2020-02-26 at 15 33 06

## Where to Start
packages/riipen-ui/src/utils/withClasses.jsx

### DIFF
--- a/packages/riipen-ui/src/utils/withClasses.jsx
+++ b/packages/riipen-ui/src/utils/withClasses.jsx
@@ -1,0 +1,34 @@
+import React from "react";
+import clsx from "clsx";
+import PropTypes from "prop-types";
+
+const withClasses = () => Component => {
+  if (process.env.NODE_ENV !== "production") {
+    if (Component === undefined) {
+      throw new Error(
+        [
+          "You are calling withStyles(styles)(Component) with an undefined component.",
+          "You may have forgotten to import it."
+        ].join("\n")
+      );
+    }
+  }
+
+  const WithClasses = React.forwardRef(function WithClasses(props, ref) {
+    const { classes, ...other } = props;
+    return (
+      <Component ref={ref} classes={[clsx("riipen", classes)]} {...other} />
+    );
+  });
+
+  WithClasses.propTypes = {
+    /**
+     * Extend the styles applied to the component.
+     */
+    classes: PropTypes.object
+  };
+
+  return WithClasses;
+};
+
+export default withClasses;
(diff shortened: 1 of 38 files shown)


packages/riipen-ui/src/utils/withClasses.jsx
```diff
@@ -0,0 +1,34 @@
...
+      );
+    }
+  }
+
+  const WithClasses = React.forwardRef(function WithClasses(props, ref) {
+    const { classes, ...other } = props;
+    return (
+      <Component ref={ref} classes={[clsx("riipen", classes)]} {...other} />
+    );
+  });
```
bolinkd: this is where the magic happens 🎆

packages/riipen-ui/src/utils/withClasses.jsx
```diff
@@ -0,0 +1,34 @@
...
+  if (process.env.NODE_ENV !== "production") {
+    if (Component === undefined) {
+      throw new Error(
+        [
+          "You are calling withStyles(styles)(Component) with an undefined component.",
+          "You may have forgotten to import it."
+        ].join("\n")
+      );
+    }
+  }
```
kradical: can we remove this? this will error when rendered anyways. And I think all components should get rendered as part of our doc site / demo?

packages/riipen-ui/src/utils/withClasses.jsx
kradical: ✨

packages/riipen-ui/src/utils/withClasses.jsx
```diff
@@ -0,0 +1,34 @@
...
+
+  const WithClasses = React.forwardRef(function WithClasses(props, ref) {
+    const { classes, ...other } = props;
+    return (
+      <Component ref={ref} classes={[clsx("riipen", classes)]} {...other} />
```
kradical: can we do `classes={["riipen", ...classes"]}` instead? I think keeping the passed in classes array mostly intact would be slightly preferable. Just in case somebody wants to do some crazy classes array shenanigans in the future.. It might be unexpected to see a single concatenated classname in that prop 🤷‍♂.

If you decided to make that change also add a defaultProp with `classes: []`!

PR #18, opened 2020-02-28
With classes fix

bolinkd:
## Description
Fixed issue with the withClasses HOC that was causing rendering issues.

## Notes
Added forwarded refs HOC

## Where to Start
packages/riipen-ui/src/utils/withClasses.jsx
packages/riipen-ui/src/utils/withForwardedRef.jsx


bolinkd:
> This all seems good to me! Why do we need withForwardedRef HoC? Is it cause the ref won't go through the withClasses component?

We just want to be more explicit about when the forwardedRef is called, if it called too frequently it gets difficult to manage as they all use the same prop. Its more of a sanity thing too make sure that we are only exposing refs that make sense... if that makes sense

PR #20, opened 2020-03-05
[PLT-4943]: Update input label, form group, and input styling

ksmort:
## Description
Updated InputLabels to be larger and gray.  Updated Input to have slightly smaller padding.  Updated FormGroup title and subtitle styling.

## Notes
Not 100% sure whether I overrode the FormGroup styles correctly.

## Screenshots
Screenshot_2020-03-05 Screenshot(3)
Screenshot_2020-03-05 Screenshot(4)

## Where to Start
/components/form
/components/input

### DIFF
--- a/packages/riipen-ui/src/components/FormGroup.jsx
+++ b/packages/riipen-ui/src/components/FormGroup.jsx
@@ -1,4 +1,6 @@
 import clsx from "clsx";
+import css from "styled-jsx/css";
+
 import PropTypes from "prop-types";
 import React from "react";
 
@@ -34,6 +36,23 @@ class FormGroup extends React.Component {
     classes: []
   };
 
+  getLinkedStyles = () => {
+    const theme = this.context;
+
+    return css.resolve`
+      .subtitle {
+        color: ${theme.palette.text.secondary};
+        font-weight: ${theme.typography.fontWeight.regular};
+      }
+
+      .title {
+        font-weight: ${theme.typography.fontWeight.medium};
+        font-size: 22px;
+        margin-bottom: ${theme.spacing(3)}px;
+      }
+    `;
+  };
+
   static contextType = ThemeContext;
 
   render() {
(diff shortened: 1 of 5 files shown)


packages/riipen-ui/src/components/InputLabel.jsx
```diff
@@ -42,15 +42,21 @@ class InputLabel extends React.Component {
...
           label {
+            color: ${theme.palette.text.secondary};
             display: inline-block;
-            margin-bottom: ${theme.spacing(1)}px;
+            font-family: ${theme.typography.body1.fontFamily};
+            font-size: font-size: 16px;
+            font-weight: ${theme.typography.body1.fontWeight};
+            letter-spacing: ${theme.typography.body1.letterSpacing};
+            line-height: 1.4;
+            margin-bottom: ${theme.spacing(3)}px;
```
ksmort: This might be too much of a margin if there is input hint.

stredger: Maybe make it changeable with props?

packages/riipen-ui/src/components/FormGroup.jsx
```diff
@@ -34,26 +36,55 @@ class FormGroup extends React.Component {
     classes: []
   };
 
+  getLinkedStyles = () => {
```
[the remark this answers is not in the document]
stredger: i think this is right, i've personally never had to do it though

packages/riipen-ui/src/components/Input.jsx
```diff
@@ -122,7 +122,7 @@ class Input extends React.Component {
             font-family: ${theme.typography.fontFamily};
             font-size: 14px;
             outline: none;
-            padding: ${theme.spacing(2)}px;
+            padding: ${theme.spacing(1.5)}px;
```
stredger: the original switch to the spacing stuff was because maribel would give us things like spacing s1 s2 s3 which were 5 10 15 etc. Dont know if we want to be doing fractions here? (unless maribel commented on the spacing)

packages/riipen-ui/src/components/FormGroup.jsx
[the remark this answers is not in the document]
bolinkd: this is correct, if you are trying to style child components and need to pass in a class you also need to pass in the parents class name or the styles wont be applied 👍

PR #21, opened 2020-03-09
Updates for new admin panel

jordanell:
## Description

1. Added `<NoSSR>` component to prevent server side rendering of children.
2. Added positive / negative buttons colors
3. Fixed bug in breadcrumbs children expecting array
4. Added "other" props to a few components.

## Notes

I ran everything through prettier so some extra files were updates as well.

### DIFF
--- a/packages/riipen-ui/src/components/Breadcrumbs.jsx
+++ b/packages/riipen-ui/src/components/Breadcrumbs.jsx
@@ -1,3 +1,4 @@
+import clsx from "clsx";
 import PropTypes from "prop-types";
 import React from "react";
 
@@ -39,15 +40,15 @@ class Breadcrumbs extends React.Component {
   static contextType = ThemeContext;
 
   render() {
-    const { children, component, separator, ...other } = this.props;
+    const { children, classes, component, separator, ...other } = this.props;
 
     const theme = this.context;
 
     return (
       <React.Fragment>
         <Typography component={component} {...other}>
-          <ol>
-            {children.map((child, index) => {
+          <ol className={clsx(classes)}>
+            {React.Children.map(children, (child, index) => {
               if (index < children.length - 1) {
                 return [
                   <li key={`child-${index}`}>{child}</li>,
--- a/packages/riipen-ui/src/components/Button.jsx
+++ b/packages/riipen-ui/src/components/Button.jsx
@@ -22,6 +22,8 @@ class Button extends React.Component {
      */
     color: PropTypes.oneOf([
       "default",
+      "negative",
+      "positive",
       "primary",
       "secondary",
       "tertiary",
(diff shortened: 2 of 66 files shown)


packages/riipen-ui/src/components/Menu.jsx
```diff
@@ -104,9 +104,8 @@ class Menu extends React.Component {
   }
 
   handleChange = (idx, event) => {
-    const { onChange, closeOnClick } = this.props;
+    const { onChange } = this.props;
     if (onChange) onChange(idx, event);
-    if (closeOnClick && event && event.type === "click") this.handleClose(idx);
   };
```
bolinkd: I think this is the only place that closeOnClick is used, 

are we saying that it is now the job of the parent component to close to menu onChange if they want to? I just dont fully know the implications for the dropdown menus on the Web app

packages/riipen-ui/src/components/ListItem.jsx
```diff
@@ -35,21 +30,16 @@ class ListItem extends React.Component {
...
+    const { children, classes, color, ...other } = this.props;
 
     const className = clsx(classes, color, "list-item");
 
     const theme = this.context;
 
     return (
       <React.Fragment>
-        <div onClick={this.handleClick} className={className}>
+        <div className={className} {...other}>
```
bolinkd: good change, this was back before I knew how/if this worked

packages/riipen-ui/src/components/Button.jsx
```diff
@@ -241,6 +243,24 @@ class Button extends React.Component {
           .contained-white:hover {
             background-color: ${theme.palette.grey[100]};
           }
+          .contained-positive {
+            background-color: ${theme.palette.positive.main};
+            color: ${theme.palette.positive.contrast};
+          }
```
bolinkd: love the colors 🌈

packages/riipen-ui/src/components/Breadcrumbs.jsx
```diff
@@ -39,15 +40,15 @@ class Breadcrumbs extends React.Component {
...
 
     const theme = this.context;
 
     return (
       <React.Fragment>
         <Typography component={component} {...other}>
-          <ol>
-            {children.map((child, index) => {
+          <ol className={clsx(classes)}>
+            {React.Children.map(children, (child, index) => {
```
bolinkd: 🚸

packages/riipen-ui/src/components/TableHeaderCell.jsx
```diff
@@ -30,13 +30,13 @@ class TableHeaderCell extends React.Component {
...
 
   render() {
-    const { align, children, classes } = this.props;
+    const { align, children, classes, ...other } = this.props;
 
     const theme = this.context;
 
     return (
       <React.Fragment>
-        <th align={align} className={clsx(classes)}>
```
bolinkd: 👍

packages/riipen-ui/src/components/Menu.jsx
jordanell: Good catch! This was actually from me screwing around previously and forgot to fix. 

I ended up just re-ordering the close and the on change triggers. The issue was that `onChange` may cause the component to unmount, then running `handleClose` would screw up.

PR #22, opened 2020-03-09
[PLT-4908] Grid Changes

bolinkd:
## Description
Changed Grid and Grid item to accept all main flex-box attributes

## Notes
Left in spacing as the grid is used in a lot of places and updating the spacing i all those places could take a long time

## Screenshots
Screen Shot 2020-03-09 at 14 02 33

## Where to Start
packages/riipen-ui/src/components/Grid.jsx
packages/riipen-ui/src/components/GridItem.jsx


packages/riipen-ui/src/components/Grid.jsx
```diff
@@ -45,14 +69,63 @@ class Grid extends React.Component {
...
      * Defines the space between grid items in the grid.
      */
-    spacing: PropTypes.oneOf(SPACINGS)
+    spacing: PropTypes.oneOf(SPACINGS),
+
+    /**
+     * The function to callback to when the grid changes sizes
+     * The callback includes the current size of the grid.
+     */
+    sizeChange: PropTypes.func
```
jordanell: I would potentially rename this one. All of our callback function props usually start with `on`. So like `onChange` or something. Maybe call this one `onResize` or `onSizeChange`?

packages/riipen-ui/src/components/Grid.jsx
```diff
@@ -45,14 +69,63 @@ class Grid extends React.Component {
...
+    spacing: 3,
+    flexDirection: "row",
+    flexWrap: "wrap",
+    sizeChange: () => {}
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      size: "lg"
```
jordanell: Is this true? What happens if the screen starts at a mobile resolution?

packages/riipen-ui/src/components/Grid.jsx
```diff
@@ -45,14 +69,63 @@ class Grid extends React.Component {
...
+      this.resizeObserver = new ResizeObserver(this.onResize);
+
+      if (this.resizeObserver && this.resizeRef.current) {
+        this.resizeObserver.observe(this.resizeRef.current);
+      }
+    } else {
+      window.addEventListener("resize", this.onResize);
+    }
+
+    this.onResize();
```
jordanell: Ah never mind my comment above!

packages/riipen-ui/src/components/Grid.jsx
bolinkd: yeah, its really just a fallback for the "this should never happen" situations

PR #23, opened 2020-03-10
[PLT-4906] Emit click event even when menu variant is not selection

bolinkd:
## Description
Added click event to menu list div that emits null event
This is useful for menus that dont have indexes to know when an option in the menu has been selected

### DIFF
--- a/packages/riipen-ui/src/components/Menu.jsx
+++ b/packages/riipen-ui/src/components/Menu.jsx
@@ -92,6 +92,8 @@ class Menu extends React.Component {
     closeOnClick: true,
     isOpen: true,
     keepOnScreen: false,
+    onChange: () => {},
+    onClose: () => {},
     popoverStyles: {},
     variant: "menu"
   };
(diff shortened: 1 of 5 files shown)


packages/riipen-ui/src/components/Menu.jsx
```diff
@@ -92,6 +92,8 @@ class Menu extends React.Component {
     closeOnClick: true,
     isOpen: true,
     keepOnScreen: false,
+    onChange: () => {},
+    onClose: () => {},
```
ksmort: Will there ever be a case where someone will want onClose to be undefined when it gets passed into Popover?  I'm kind of wary with empty functions, I usually like making them required as a prop or just undefined, but that's just me 😛

bolinkd: hmm, I can see your point. I was just trying to clean up the `if (onClose) onClose()` repeats that were appearing everywhere. Ill revert it for now as there is no reason other than that I was going to change it

PR #25, opened 2020-03-10
Fix label font size and form group styling

ksmort:
## Description
Fix typo with InputLabel font size and add more specificity to form group styles so that styles actually get applied.

## Where to Start
FormGroup and InputLabel components.

### DIFF
--- a/packages/riipen-ui/package.json
+++ b/packages/riipen-ui/package.json
@@ -1,6 +1,6 @@
 {
   "name": "riipen-ui",
-  "version": "0.0.83",
+  "version": "0.0.84",
   "private": false,
   "author": {
     "name": "Riipen",
--- a/packages/riipen-ui/package.json
+++ b/packages/riipen-ui/package.json
@@ -1,6 +1,6 @@
 {
   "name": "riipen-ui",
-  "version": "0.0.82",
+  "version": "0.0.83",
   "private": false,
   "author": {
     "name": "Riipen",
--- a/packages/riipen-ui/src/components/FormGroup.jsx
+++ b/packages/riipen-ui/src/components/FormGroup.jsx
@@ -40,12 +40,12 @@ class FormGroup extends React.Component {
     const theme = this.context;
 
     return css.resolve`
-      .subtitle {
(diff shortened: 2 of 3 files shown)


packages/riipen-ui/src/components/FormGroup.jsx
```diff
@@ -40,12 +40,12 @@ class FormGroup extends React.Component {
     const theme = this.context;
 
     return css.resolve`
-      .subtitle {
+      p.subtitle {
```
bolinkd: what was the reason we needed to add this? just wondering. I just believe it would work either way

packages/riipen-ui/src/components/InputLabel.jsx
```diff
@@ -58,7 +58,7 @@ class InputLabel extends React.Component {
             color: ${theme.palette.text.secondary};
             display: inline-block;
             font-family: ${theme.typography.body1.fontFamily};
-            font-size: font-size: 16px;
```
bolinkd: opps, missed that one myself 🙈

packages/riipen-ui/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "riipen-ui",
-  "version": "0.0.83",
+  "version": "0.0.84",
```
bolinkd: do you need me to deploy the new version when it is done?

packages/riipen-ui/src/components/FormGroup.jsx
ksmort: I'm having an issue where the .title and .subtitle styling are getting overridden by .h3 and .root styling when applied to web.  You're right though, I should do some more digging and figure out the root cause of it.

packages/riipen-ui/package.json
ksmort: That would be awesome

packages/riipen-ui/src/components/InputLabel.jsx
ksmort: Mondays 🙈

PR #30, opened 2020-03-18
Add expandable row to tableGenerator

stredger:
## Description
This is what i eventually landed on. Basically just pass a function to determine which rows to display expanded or not. Originally i had it as an attribute on the entity, but dont think modifying the entity is ideal

## Screenshots
Screen Shot 2020-03-18 at 4 37 45 PM

### DIFF
--- a/packages/riipen-ui/src/components/TableGenerator.jsx
+++ b/packages/riipen-ui/src/components/TableGenerator.jsx
@@ -88,7 +88,8 @@ class TableGenerator extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      isMobile: false
+      isMobile: false,
+      hoverIdx: null
     };
 
     this.updateWindowDimensions = debounce(this.updateWindowDimensions);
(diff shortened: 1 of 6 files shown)


packages/riipen-ui/src/components/TableGenerator.jsx
```diff
@@ -176,23 +193,34 @@ class TableGenerator extends React.Component {
...
+                    </TableCell>
+                  ))}
+                </TableRow>
+                {isExpandable && expandRow(row) && (
+                  <TableRow>
+                    <TableCell colSpan={columns.length}>
+                      {expandedNode(row)}
+                    </TableCell>
+                  </TableRow>
+                )}
```
bolinkd: I dont have time to pull this locally at the moment so just something to look into for the morning, I think this wont work correctly with the hover prop. According to the designs the main row and hovered row should be highlighted when moused over. I dont know if this is possible as I dont know exactly how moving the cells into the parent row would work (would it wrap the row?) Something to look into for now

stredger: yea the expanded stuff does not get "hovered" when you hover over the row. i guess for consistency i should add it

packages/riipen-ui/src/components/TableRow.jsx
```diff
@@ -25,23 +30,48 @@ class TableRow extends React.Component {
...
+      onMouseOut
+    } = this.props;
 
     const theme = this.context;
 
     return (
       <React.Fragment>
-        <tr className={clsx(classes)}>{children}</tr>
+        <tr
+          className={clsx(classes, forceHover && "forceHover")}
```
bolinkd: dont quote me on this, but I believe you can replace `forceHover && "forceHover"` with `{ forceHover }`

PR #31, opened 2020-03-19
Add tooltip component

ksmort:
## Description
Create Tooltip component based on Popover.  This Tooltip component has little arrows based on the contentPosition prop.  There are also colour options.

## Notes
Definitely need help on the description for the tooltips, I'm very bad at that!

TODO: animation transitions, other things I might have missed

## Screenshots
tooltip
tooltip2
tooltip3
tooltip5
tooltip6
tooltip4

## Where to Start
components/Tooltip

### DIFF
--- a/packages/riipen-ui-docs/src/pages.js
+++ b/packages/riipen-ui-docs/src/pages.js
@@ -116,6 +116,10 @@ const pages = [
         pathname: "/components/table"
       },
       {
+        name: "Tooltip",
+        pathname: "/components/tooltip"
+      },
+      {
         name: "Typography",
         pathname: "/components/typography"
       }
@@ -203,6 +207,10 @@ const pages = [
       {
         name: "Toolbar",
         pathname: "/components-api/toolbar"
+      },
+      {
+        name: "Tooltip",
+        pathname: "/components-api/tooltip"
       },
       {
         name: "Typography",
--- a/packages/riipen-ui-docs/src/pages/components/tooltip/tooltip.md
+++ b/packages/riipen-ui-docs/src/pages/components/tooltip/tooltip.md
@@ -2,7 +2,7 @@
 
 <p class="description">The Tooltip will create a place for information to appear over top existing data</p>
 
-The Tooltip wraps the Popover component. You can pass Popover props into it.
+The Tooltip wraps the Popover component. You can pass Popover props into it. It uses the contentPosition prop to determine the arrow style.
 
 The Tooltip has a variety of colours to choose from.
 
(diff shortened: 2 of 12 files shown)


packages/riipen-ui-docs/src/pages/components/tooltip/tooltip.md
```diff
@@ -0,0 +1,21 @@
+# Tooltip
+
+<p class="description">The Tooltip will create a place for information to appear over top existing data</p>
```
ksmort: Could use help on making these sound good. 😬

packages/riipen-ui/src/components/Tooltip.jsx
```diff
@@ -0,0 +1,429 @@
...
+import ThemeContext from "../styles/ThemeContext";
+
+import withClasses from "../utils/withClasses";
+
+import Popover from "./Popover";
+
+class Tooltip extends React.Component {
+  static propTypes = {
+    /**
+     * The tooltip content.
```
jordanell: I would maybe explain this one a bit more, it's a bit confusing given the documentation for the `tooltip` prop. I would maybe say "The content to trigger the tooltip with". or something.

packages/riipen-ui-docs/src/pages.js
```diff
@@ -115,6 +115,10 @@ const pages = [
         name: "Table",
         pathname: "/components/table"
       },
+      {
+        name: "Tooltip",
+        pathname: "/components/tooltip"
+      },
```
bolinkd: reminder to one day automate this... brb, creating ticket

packages/riipen-ui/src/components/Tooltip.jsx
```diff
@@ -0,0 +1,427 @@
...
+     * The component used for the root node.
+     * Either a string to use a DOM element or a component.
+     */
+    component: PropTypes.elementType,
+
+    /**
+     * The location to attach the anchor to on the content element.
+     * Also used to determine the tooltip arrow placement/direction.
+     */
+    contentPosition: PropTypes.object,
```
bolinkd: I would consider changing this to be just position with values ["above", "below", "left", "right"] and the calculate the anchor and content position based on that. I dont think we need that much extensibility and it would simplify the API considerably. Thoughts?

ksmort: That does sound nice!  Would we be able to combine the positions to have things like top-right, bottom-left etc?  

I have a use case in design where I want the arrow in the top left corner, which is why I think I went with contentPosition, but happy to update!

bolinkd: yeah, we could do the 8 positions and I would suggest setting the anchor and content positions accordingly so that they line up properly.
ie: 
// above-left
anchorPosition: { vertical: "top", horizontal: "left" }
contentPosition: { vertical: "bottom", horizontal: "right" }

// below
anchorPosition: { vertical: "bottom", horizontal: "center" }
contentPosition: { vertical: "top", horizontal: "center" }

ksmort: Added the positions prop, used top/bottom instead of above/below because it was easy to parse 🙈 Let me know what you think!

PR #32, opened 2020-03-20
Popover changes: fullWidth prop, handleFragmentChildren()

ericauv:
## Description
* Need option to have popover be full-width for mobile popover menus
* For mobile account dropdown https://projects.invisionapp.com/d/main/#/console/19399905/407497871/inspect
## Screenshots
Screen Shot 2020-03-20 at 2 31 30 PM

## Where to Start
/components/Popover.js


bolinkd:
you should be able to do this with the props provided, lets talk after social interaction

packages/riipen-ui/src/components/Popover.jsx
```diff
@@ -228,6 +228,11 @@ class Popover extends React.Component {
       }
     }
 
+    // Handle full-width menu
+    if (anchorPosition.horizontal === "full-width") {
+      left = 0;
+    }
+
     this.setState(
```
bolinkd: I would consider making a fullWidth prop, and the if that is set doing a separate calculation at the beginning of this function. That way the logic is separated and it would be easy to remove later if we can fix the issue

packages/riipen-ui/src/components/Popover.jsx
```diff
@@ -228,6 +235,11 @@ class Popover extends React.Component {
       }
     }
 
+    // Handle fullWidth menu
+    if (fullWidth) {
+      left = 0;
```
ericauv: Added to end of function, if added to beginning, would need to wrap each of the changes to `left` throughout the function in an `if(!fullWidth){}`, since still need the `top` changes throughout the function to get `top` position of `Popover`

Feel this is cleaner / more easily changed

bolinkd: that's perfectly reasonable

PR #34, opened 2020-03-24
add handleFragmentChildren

ericauv:
## Description
* If a MenuList's children were wrapped in a React.Fragment, the 'color' prop was not being passed as expected
* Add `handleFragmentChildren()` to `getListItems`
## Where to Start
/components/MenuList

### DIFF
--- a/packages/riipen-ui/src/components/MenuList.jsx
+++ b/packages/riipen-ui/src/components/MenuList.jsx
@@ -106,7 +106,7 @@ class MenuList extends React.Component {
   }
 
   keyDownHandler = event => {
-    const { selectedIndex, children } = this.props;
+    const { children, selectedIndex } = this.props;
     const nextEvents = ["ArrowDown", "ArrowRight"];
     const prevEvents = ["ArrowUp", "ArrowLeft"];
     const enterEvent = "Enter";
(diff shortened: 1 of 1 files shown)


packages/riipen-ui/src/components/MenuList.jsx
```diff
@@ -69,26 +69,31 @@ class MenuList extends React.Component {
...
-        key: idx,
-        color: this.props.color
-      };
-      if (variant === "selection") {
-        newProps = Object.assign(newProps, {
-          onClick: this.handleClick(child, idx)
-        });
-        if (idx === activeItemIndex) {
+    return React.Children.map(
+      this.handleFragmentChildren(children),
```
ericauv: This line is the only thing that changed in this function

packages/riipen-ui/src/components/MenuList.jsx
```diff
@@ -153,6 +158,17 @@ class MenuList extends React.Component {
...
+    let handledChildren = children;
+
+    if (!Array.isArray(handledChildren)) {
+      if (handledChildren.type.toString().includes("react.fragment")) {
+        handledChildren = this.handleFragmentChildren(children.props.children);
+      }
+    }
+    return handledChildren;
+  };
+
```
bolinkd:
```suggestion
  handleFragmentChildren = children => {
      return children.type === React.Fragment
            ? children.props.children
            : children;
  };

```

PR #35, opened 2020-03-25
Update tooltip position, keep tooltip on screen, add fade animation

ksmort:
## Description
Update Tooltip to add fade in animation and keep tooltip on screen.  Update positions so that they are actually top-right = tooltip displays on top right (it was the opposite which was confusing).

Update example wording.

## Screenshots
tooltip

## Where to Start
Tooltip

### DIFF
--- a/packages/riipen-ui/src/components/Tooltip.jsx
+++ b/packages/riipen-ui/src/components/Tooltip.jsx
@@ -108,6 +108,17 @@ class Tooltip extends React.Component {
         width: max-content;
       }
 
+      .popover {
+        opacity: 0;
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/Tooltip.jsx
```diff
@@ -175,62 +186,62 @@ class Tooltip extends React.Component {
         position: absolute;
       }
 
-      .popover.top {
+      .popover.bottom {
```
ksmort: Future TODO: make DRY up this css.

PR #40, opened 2020-03-31
Make buttons the same size

ksmort:
## Description
-Added line height so that buttons with/without icons are the same height
-Decreased the padding of outlined buttons by 1 px to accommodate the 1 px border
-Updated the spacing to some buttons in the examples

## Notes
There's still one really pesky thing going on with buttons with icons: buttons with an iconStart prop display slightly higher than other buttons.  I _think_ something is going on with the flex container, but still can't quite pinpoint the behaviour.  See attached screenshot and also https://ui.riipen.com/components/button/#icons.

## Screenshots
Screenshot_2020-03-31 Screenshot
Screenshot_2020-03-31 Screenshot(1)
Screenshot_2020-03-31 Screenshot(2)

### DIFF
--- a/packages/riipen-ui/src/components/Button.jsx
+++ b/packages/riipen-ui/src/components/Button.jsx
@@ -320,6 +320,16 @@ class Button extends React.Component {
             background-color: ${theme.palette.negative.main};
           }
 
+          .outlined.small {
+            padding: calc(${theme.spacing(1)}px - 1px)
+              calc(${theme.spacing(2)}px - 1px);
+          }
+
+          .outlined.large {
+            padding: calc(${theme.spacing(3)}px - 1px)
+              calc(${theme.spacing(5)}px - 1px);
+          }
+
           .outlined {
             background-color: transparent;
             border: 1px solid rgba(0, 0, 0, 0.23);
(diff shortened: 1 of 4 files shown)


ksmort:
Updated the implementation, not sure if it's great but I think it's a bit better than the first way?

packages/riipen-ui/src/components/Button.jsx
```diff
@@ -322,11 +317,20 @@ class Button extends React.Component {
...
+          .outlined::after {
+            border-radius: ${theme.shape.borderRadius.md};
+            border: 1px solid rgba(0, 0, 0, 0.23);
+            bottom: 0;
+            content: "";
+            left: 0;
+            position: absolute;
+            right: 0;
+            top: 0;
+          }
```
stredger, line 333: Does this just make the border not take up the extra pixel?

ksmort: Yeah, it creates a pseudo-element and then applies the border to that, so it doesn't take up the extra pixel

ksmort:
> Looks good! Just curious why you changed it?

Looking it over I thought calculating padding - 1 looked a little too hacky, idk 🤷‍♀ 😛

PR #44, opened 2020-04-17
Update tooltip visibility

ksmort:
## Description
Move passing props to the Popover in the Tooltip component to allow for some overrides.  Also fix an issue with whitespace and the tooltip arrows.

## Where to Start
/components/Tooltip

### DIFF
--- a/packages/riipen-ui/src/components/Tooltip.jsx
+++ b/packages/riipen-ui/src/components/Tooltip.jsx
@@ -113,6 +113,7 @@ class Tooltip extends React.Component {
       .popover {
         border-radius: 2px;
         color: ${theme.palette.common.white};
+        display: none;
       }
 
       @keyframes fade {
@@ -128,6 +129,7 @@ class Tooltip extends React.Component {
       .popover.show {
         animation: fade ${theme.transitions.duration.short}ms
           ${theme.transitions.easing.easeIn};
+        display: block;
       }
 
       /* Sizes */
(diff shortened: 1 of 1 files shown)


bolinkd:
What exactly is the reason for this change? Does it have something to do with transitions? I havent tried applying transitions to popovers yet

packages/riipen-ui/src/components/Tooltip.jsx
```diff
@@ -197,6 +197,7 @@ class Tooltip extends React.Component {
         border-width: ${theme.spacing(2)}px;
         content: "";
         position: absolute;
+        white-space: normal;
```
ksmort, line 200: Set whitespace here so that the tooltip arrows are not moved around unexpectedly

packages/riipen-ui/src/components/Tooltip.jsx
```diff
@@ -466,11 +467,12 @@ class Tooltip extends React.Component {
           horizontal: contentHorizontal,
           vertical: contentVertical
         }}
-        isOpen={open}
         anchorEl={this.tooltipRootRef.current}
+        isOpen={open}
         keepOnScreen
         lockScroll={false}
         onClose={this.handleClose}
+        {...other}
```
ksmort, line 475: allows for some overrides if needed

ksmort:
> What exactly is the reason for this change? Does it have something to do with transitions? I havent tried applying transitions to popovers yet

The transitions are working okay, it was an attempt to fix some issues with the new tooltip and the interactive tooltip in web, but I think I'll do it differently in web instead.

PR #46, opened 2020-04-28
[PLT-5158] Added inverted button type

bolinkd:
## Description
Added inverted button type for Search Page Header

## Screenshots
Screen Shot 2020-04-28 at 15 57 29

## Where to Start
packages/riipen-ui/src/components/Button.jsx

### DIFF
--- a/packages/riipen-ui/src/components/Button.jsx
+++ b/packages/riipen-ui/src/components/Button.jsx
@@ -448,13 +448,6 @@ class Button extends React.Component {
             color: ${theme.palette.common.white};
             background-color: ${theme.palette.tertiary.main};
           }
-          .inverted-white {
-            color: ${theme.palette.common.black};
-          }
-          .inverted-white:hover {
-            color: ${theme.palette.common.white};
-            background-color: ${theme.palette.common.black};
-          }
           .inverted-positive {
             color: ${theme.palette.positive.main};
           }
--- a/packages/riipen-ui/src/components/Button.jsx
+++ b/packages/riipen-ui/src/components/Button.jsx
@@ -409,7 +409,7 @@ class Button extends React.Component {
           }
           .inverted::after {
             border-radius: ${theme.shape.borderRadius.md};
-            border: 1px solid rgba(0, 0, 0, 0.23);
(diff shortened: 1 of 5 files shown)


packages/riipen-ui/src/components/Button.jsx
```diff
@@ -400,6 +401,88 @@ class Button extends React.Component {
...
+          .inverted-tertiary {
+            color: ${theme.palette.tertiary.main};
+          }
+          .inverted-tertiary:hover::before {
+            background-color: ${theme.palette.tertiary.main};
+          }
+          .inverted-tertiary::after {
+            border-color: ${theme.palette.tertiary.main};
+          }
+          .inverted-white {
```
ericauv, line 458:
```suggestion
          .inverted-black {
```
To maintain consistency of naming, I think we should continue using the `color` as `${color}` in `inverted-${color}`

packages/riipen-ui/src/components/Button.jsx
```diff
@@ -400,6 +401,88 @@ class Button extends React.Component {
...
+            color: ${theme.palette.common.black};
+          }
+          .inverted-white:hover::before {
+            background-color: ${theme.palette.grey[100]};
+          }
+          .inverted-white::after {
+            border-color: ${theme.palette.common.black};
+          }
```
ericauv, line 466:
```suggestion
          .inverted-black {
            color: ${theme.palette.common.black};
          }
          .inverted-black:hover::before {
            background-color: ${theme.palette.grey[100]};
          }
          .inverted-black::after {
            border-color: ${theme.palette.common.black};
          }
```

ericauv: Would need to add `black` to `propTypes:{color}` also

ericauv: / For `inverted-white`, I think it should have `{backgroundColor: black, color: white}`

bolinkd: I dont think the use case for this is very large, I am going to remove it for now and let someone build it when the use case presents itself

PR #47, opened 2020-04-28
[PLT-5048]: Update button icon hover and default colour

ksmort:
## Description
Update ButtonIcon's default colour.  Also update default button to be orange on hover.  Added "muted" colour so that there is a gray button that hovers on gray.

## Screenshots
icon

## Where to Start
components/Button

### DIFF
--- a/packages/riipen-ui/src/components/ButtonIcon.jsx
+++ b/packages/riipen-ui/src/components/ButtonIcon.jsx
@@ -90,7 +90,7 @@ class ButtonIcon extends React.Component {
           .root {
             background-color: transparent;
             border: 0;
-            color: ${theme.palette.text.primary};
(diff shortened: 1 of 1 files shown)


packages/riipen-ui/src/components/ButtonIcon.jsx
```diff
@@ -22,6 +22,7 @@ class ButtonIcon extends React.Component {
      */
     color: PropTypes.oneOf([
       "default",
+      "muted",
```
ksmort, line 25: Just used the Text component wording, but maybe this should be "grey" or something else?

stredger: i think this is fine

PR #48, opened 2020-05-08
[PLT-5120]: Add outline prop to ButtonIcon

ericauv:
Relates to [[PLT-5120]](https://riipen.atlassian.net/browse/PLT-5120)

## Description
* Add outline prop to `ButtonIcon` to improve accessibility
* Required for https://github.com/riipen/web/pull/2475
* Bump package version

## Notes
* Should likely have this default to true so that all `ButtonIcon` components are accessible, but have put as false for now until discussion determines we want all `ButtonIcon` components to have outline by default

## Screenshots
outline

### DIFF
--- a/packages/riipen-ui/src/components/ButtonIcon.jsx
+++ b/packages/riipen-ui/src/components/ButtonIcon.jsx
@@ -41,6 +41,11 @@ class ButtonIcon extends React.Component {
     disabled: PropTypes.bool,
 
     /**
+     * If `true`, the button will have an outline. If `false` the button will not have an outline when focussed.
+     */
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/ButtonIcon.jsx
```diff
@@ -40,6 +40,11 @@ class ButtonIcon extends React.Component {
      */
     disabled: PropTypes.bool,
 
+    /**
+     * If `true`, the button will have an outline. If `false` the button will not have an outline when focussed.
```
jordanell, line 44: Maybe add "when focused" to the first sentence.

PR #49, opened 2020-05-12
Rich text editor

ksmort:
## Description
Update scripts to auto-generate subdirectories.  Update menu to display subdirectories (see notes).
Add rich text editor to UI kit.  Add editor components into editor subdirectory.
-move over editor util functions from web so that can be exported as ../utils/editor so that the draft-js stuff all lives together

## Notes
Open to other ways to display subdirectories in menu.
There are a bunch of tests in web for these components that would be awesome to copy-pasta/convert over eventually.
-some potential code spaghetti in editor/Editor, see comment in that file

## Screenshots
Screenshot_2020-05-12 Screenshot(1)
Screenshot_2020-05-12 Screenshot(2)

## Where to Start
/components/rich-text-editor


packages/riipen-ui-docs/src/pages/getting-started/usage/usage.md
```diff
@@ -2,7 +2,7 @@
 
 <p class="description">Get started with React and Riipen-UI in no time.</p>
 
-Material-UI components work in isolation. **They are self-supporting**, and will only
+Riipen-UI components work in isolation. **They are self-supporting**, and will only
```
ksmort, line 5: typo

packages/riipen-ui/src/components/editor/BlockStyleControls.jsx
```diff
@@ -0,0 +1,105 @@
+import clsx from "clsx";
+import PropTypes from "prop-types";
+import React from "react";
+
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import {
+  faListOl,
+  faListUl,
+  faQuoteLeft
+} from "@fortawesome/free-solid-svg-icons";
```
ksmort, line 10: We originally used the pro icons in the original RTE, not sure how to handle that.

packages/riipen-ui/src/components/editor/Editor.jsx
```diff
@@ -0,0 +1,586 @@
...
+  }
+
+  componentDidMount() {
+    const { initialValue } = this.props;
+
+    // Sets editor state after mount (because SSR)
+    const editorState = initialValue
+      ? EditorState.createWithContent(convertFromHTML(initialValue), decorator)
+      : EditorState.createEmpty(decorator);
+    this.onChange(editorState);
```
ksmort, line 145: Handle server-side rendering

packages/riipen-ui/src/components/editor/BlockStyleControls.jsx
```diff
@@ -0,0 +1,105 @@
...
+            onClick={this.onControlClick(style)}
+            width={width}
+          />
+        ))}
+      </div>
+    );
+  }
+}
+
+export default BlockStyleControls;
```
ksmort, line 105: @bolinkd Do I need to have withClasses here even if there is no css?

bolinkd: I am actually working towards removing the withClasses in the future, it was supposed to add an additional selector for the component but it ended up not working due to name mutilation that happens on production builds
its supposed to add .riipen and .riipen-{componentName} but it ended up always generating .riipen-t in production as the componentName gets removed in production builds.

packages/riipen-ui/src/components/editor/Editor.jsx
```diff
@@ -0,0 +1,595 @@
...
+                key={`control-${index}`}
+                className={clsx([linkedStyles.className, "controlRow"])}
+              >
+                {control}
+              </div>
+            ))}
+        </div>
+      </React.Fragment>
+    );
+  };
```
ksmort, line 529: These functions are actually not used by this component.  In web they were here and then components such as Messages/Form/New and ImageControls used a ref to get them.  I left them here for now which kind of feels spaghetti.  We could move them into web, but then the draft-js stuff doesn't all live together.  Input appreciated!

packages/riipen-ui/src/utils/editor/getEntitiesSelection.js
```diff
@@ -0,0 +1,77 @@
+import { findLinkEntities } from "./findEntities";
```
ksmort, line 1: Throw this and getLinksSelection so that they live with the RTE

packages/riipen-ui-docs/src/pages.js
```diff
@@ -160,6 +164,35 @@ const pages = [
         name: "Drawer",
         pathname: "/components-api/drawer"
       },
+      {
+        name: "Editor",
+        children: [
```
jordanell, line 169: Nice.

packages/riipen-ui-docs/src/pages/getting-started/usage/usage.md
jordanell: LOL. Guess where I stole most of the architecture code from this repo from?

packages/riipen-ui/src/components/editor/BlockStyleControls.jsx
jordanell: I think this is actually better. Helps consolidate all of our iconography.

packages/riipen-ui/src/components/editor/ControlButton.jsx
```diff
@@ -0,0 +1,97 @@
...
+     * Function to execute on control button click.
+     *
+     */
+    onClick: PropTypes.func.isRequired,
+
+    /**
+     * The width of the control button.
+     *
+     */
+    width: PropTypes.oneOf(["small", "medium", "large"])
```
jordanell, line 32: I think we should change this prop name to `size` instead of `width`. It would be more inline with our naming conventions on other components that have small / medium / large options.

packages/riipen-ui/src/components/editor/Editor.jsx
jordanell: I am find with leaving them in here as they are basically part of this component's public API. Maybe adding a section to the component documentation that shows what available methods there are?

packages/riipen-ui/src/components/editor/Link.jsx
```diff
@@ -0,0 +1,53 @@
+import { ContentState } from "draft-js";
+import PropTypes from "prop-types";
+import React from "react";
+
+import ThemeContext from "../../styles/ThemeContext";
+import withClasses from "../../utils/withClasses";
+
+class RteLink extends React.Component {
+  static propTypes = {
+    // draft-js CompositeDecorator
```
jordanell, line 10: Can probably remove this comment?

packages/riipen-ui/src/components/editor/Editor.jsx
ksmort: Good idea, will do!

ksmort:
> Looking good overall. Left some comments about naming conventions.
> 
> I did have a concern about all the new `utils` files. I am wondering if all of these functions should just live inside of their appropriate components, or if they aren't used in a component, if we should remove them entirely. I sort of like the idea of the ui kit being solely focus on ui and minimal functionality, and if apps, need to do more advanced stuff then to offload that work to them. Thoughts?

Yeah, that sounds good!

packages/riipen-ui/src/components/editor/Editor.jsx
```diff
@@ -0,0 +1,638 @@
...
+
+    // Checks if text is highlighted to either insert/replace
+    if (selectionState.isCollapsed()) {
+      newContentState = Modifier.insertText(contentState, selectionState, text);
+    } else {
+      newContentState = Modifier.replaceText(
+        contentState,
+        selectionState,
+        text
+      );
```
ksmort, line 399: Fixes error being thrown because insertText requires text not to be highlighted

packages/riipen-ui/src/components/editor/Editor.jsx
```diff
@@ -0,0 +1,638 @@
...
+
+    /**
+     * Position of controls for the editor
+     */
+    controlPosition: PropTypes.oneOf(["top", "bottom"]),
+
+    /**
+     * The decorator for the editor.
+     */
+    decorator: PropTypes.shape({ type: PropTypes.oneOf([CompositeDecorator]) }),
```
ksmort, line 120: Moved the decorator out so that devs need to pass custom decorators.  The only place this makes a difference so far is with links, but we'll pass a decorator in so should be no problemo.

PR #50, opened 2020-05-13
[PLT-5122] Added accessibility attributes to tabs and checkboxes

bolinkd:
## Description
Fixed issue where menu items were not keyboard accessible
Fixed issue where outline for checkboxes was not appearing

## Notes
Learned lots about a11y attributes... also, turns out you use space to toggle checkboxes... not enter

## Screenshots
Menus:
Screen Shot 2020-05-13 at 15 33 47

Checkboxes:
Screen Shot 2020-05-13 at 15 34 29

### DIFF
--- a/packages/riipen-ui/src/components/Tab.jsx
+++ b/packages/riipen-ui/src/components/Tab.jsx
@@ -71,10 +71,13 @@ class Tab extends React.Component {
 
   static contextType = ThemeContext;
 
-  handleClick = e => {
+  handleChange = e => {
     const { onClick, value } = this.props;
+    if (e.type === "click" || (e.type === "keydown" && e.key === "Enter")) {
+      onClick(e, value);
+    }
 
-    return onClick(e, value);
+    return;
   };
 
   render() {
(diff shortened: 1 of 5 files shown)


packages/riipen-ui/src/components/Tab.jsx
```diff
@@ -104,7 +107,14 @@ class Tab extends React.Component {
 
     return (
       <React.Fragment>
-        <div className={className} onClick={this.handleClick}>
+        <div
+          tabIndex="0"
```
jordanell, line 111: Is this correct? Would this always put it as the first tab location even if it's in the middle of the page?

bolinkd: took a bit of time looking into how this works yesterday actually

tabIndex -1: The item can be focused but must be done by a script, using the tab key wont work
tabIndex 0: The item can be focused by script or tab key
Any item with the same tabIndex (default is 0 for most things that can be focused) will tab in order from top to bottom
If an item has a higher tab index it follows the above rule but with higher priority.
So you could have everything on your page have a tab index of 1 and focus would work the same as it does now. 

Hopefully I explained it well @jordanell

bolinkd: oh idea how to explain, it works in the same way as zIndex

PR #52, opened 2020-06-02
[PLT 5370 / RB-1571]: Fix undefined event error

ericauv:
[RB-1571](https://rollbar.com/riipen/web/items/1571/)
[[PLT 5370]](https://riipen.atlassian.net/browse/PLT-5370)

## Description
* Error occurs when `stopPropagation()` is called on `event` when `event` is `undefined`
* Pass `event` from `ListItem` to `handleClick` in `MenuItem`

### DIFF
--- a/packages/riipen-ui/src/components/MenuItem.jsx
+++ b/packages/riipen-ui/src/components/MenuItem.jsx
@@ -153,7 +153,10 @@ class MenuItem extends React.Component {
 
     const handleClick = event => {
       if (disabled) {
-        event.stopPropagation();
+        if (event) {
+          event.preventDefault();
+          event.stopPropagation();
+        }
       }
 
       if (onClick && !disabled) onClick(event);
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/MenuItem.jsx
```diff
@@ -152,7 +152,8 @@ class MenuItem extends React.Component {
     );
 
     const handleClick = event => {
-      if (disabled) {
+      if (disabled && event) {
```
ksmort, line 155: It seems a bit strange that only sometimes an event is passed to the handleClick function?  Maybe something is going on in ListItem where it gets passed?

ericauv: it looks like event is always `undefined` -- I think it's because we are calling `handleClick` on the `onClick` of `ListItem`, and the `ListItem` component is wrapped in a `<React.Fragment>`?

I think this error hasn't popped up before because we don't seem to actually use the event in any of the `onClick` functions that are passed to `MenuItem`

I could only find `onClick` props that were functions that didn't use the `event` at all

ksmort: Nice sleuthing! 🕵️

PR #54, opened 2020-07-08
[PLT-5265] Popover, Tooltip improvements

ericauv:
[[PLT-5265]](https://riipen.atlassian.net/browse/PLT-5265)

## Description
* Fix bug with `Popover` where initial position would be off by (properContentRectWidth/2 - 21px)
* Add `keydown` handler to `Popover` / `Tooltip`
* Add `closeOnParentScroll` prop to `Popover` --> closes Popover when scrollable Parent container is scrolled
* Add `getClosestScrollableParent` util
* Add ability for `Tooltip` open/close state to be controlled by props

## Notes
* Required for https://github.com/riipen/web/pull/2585

### DIFF
--- a/packages/riipen-ui/src/components/Popover.jsx
+++ b/packages/riipen-ui/src/components/Popover.jsx
@@ -40,7 +40,7 @@ class Popover extends React.Component {
      * Whether the popover should be closed when its closest scrollable parent is scrolled.
      * [Default] true
      */
-    closeWhenParentScrolled: PropTypes.bool,
+    closeOnScrolled: PropTypes.bool,
 
     /**
      * The type of element to use at the root
@@ -65,7 +65,7 @@ class Popover extends React.Component {
     /**
      * Function call to handle keydown events,
      */
-    handleKeyDown: PropTypes.func,
+    onKeyDown: PropTypes.func,
 
     /**
      * Whether to keep the popout on screen when the anchor element scrolls off
@@ -95,7 +95,7 @@ class Popover extends React.Component {
   };
 
   static defaultProps = {
-    closeWhenParentScrolled: true,
+    closeOnScrolled: true,
     component: "span",
     fullWidth: false,
     isOpen: false,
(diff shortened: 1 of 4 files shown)


packages/riipen-ui/src/components/Popover.jsx
```diff
@@ -126,21 +143,32 @@ class Popover extends React.Component {
...
+
     if (
       anchorChange ||
       openChange ||
       contentPositionChange ||
       anchorPositionChange
     ) {
-      this.setPositioningStyle();
+      // need to use requestAnimationFrame in order for popover to get proper position on first open
+      window.requestAnimationFrame(this.setPositioningStyle);
```
ericauv, line 171: this fixes initial positioning `Popover` bug, but only when the `Popover` is opened normally (with MouseEnter / Click), need changes on `L151` to fix bug when `Popover` is opened via 'changing tab' in an `InteractiveTooltip`

packages/riipen-ui/src/components/Popover.jsx
```diff
@@ -251,6 +282,42 @@ class Popover extends React.Component {
     );
   };
 
+  manageScrollableParentEventListeners(remove = true) {
```
ericauv, line 285: for adding/removing event listeners on the closest scrollable parent

packages/riipen-ui/src/components/Popover.jsx
```diff
@@ -280,6 +347,11 @@ class Popover extends React.Component {
       this.handleClose();
     } else if (event.type === "keydown" && event.key === "Escape") {
       this.handleClose();
+    } else if (event.type === "keydown" && this.props.handleKeyDown) {
+      // stop this event from triggering other keydown listeners
+      event.stopImmediatePropagation();
```
ericauv, line 352: need to stop this event from triggering other keydown listeners, otherwise when 'changing tab' in `InteractiveTooltip`, all `Popover` `keydown` listeners fire, causing the tab to change `numberOfPopovers` times

packages/riipen-ui/src/components/Popover.jsx
```diff
@@ -55,6 +62,11 @@ class Popover extends React.Component {
      */
     isOpen: PropTypes.bool,
 
+    /**
+     * Function call to handle keydown events,
+     */
+    handleKeyDown: PropTypes.func,
```
ksmort, line 68: Could maybe call onKeyDown to match other on(thing happens) function calls, but no biggie if not!

packages/riipen-ui/src/components/Tooltip.jsx
```diff
@@ -42,11 +42,23 @@ class Tooltip extends React.Component {
      */
     component: PropTypes.elementType,
 
+    /**
+     * Function to call on tooltip keydown.
+     */
+    handleKeyDown: PropTypes.func,
```
ksmort, line 48: same here

packages/riipen-ui/src/utils/container.js
```diff
@@ -17,4 +17,22 @@ export function getContainer(node) {
   return doc.defaultView || window;
 }
 
+/**
+ * Get closest parent that is scrollable
+ * @param {element} node - An element in the DOM
+ * @return {string} The closest scrollable parent view element that contains the element
+ */
+export function getClosestScrollableParent(node) {
```
bolinkd, line 25: 👍

packages/riipen-ui/src/components/Popover.jsx
bolinkd: agree with @ksmort, the general scheme to follow, 
- if prop: on{Event}
- if function in component: handle{Event}

packages/riipen-ui/src/components/Popover.jsx
```diff
@@ -35,6 +36,12 @@ class Popover extends React.Component {
      */
     classes: PropTypes.array,
 
+    /**
+     * Whether the popover should be closed when its closest scrollable parent is scrolled.
+     * [Default] true
+     */
+    closeWhenParentScrolled: PropTypes.bool,
```
bolinkd, line 43: This could just be closeOnScrolled or something similar. The idea is that it gets closed when it moves... original just seems a little wordy to me. Not a huge deal though

packages/riipen-ui/src/components/Popover.jsx
```diff
@@ -83,9 +95,10 @@ class Popover extends React.Component {
   };
 
   static defaultProps = {
+    closeOnScrolled: true,
     component: "span",
     fullWidth: false,
-    isOpen: true,
+    isOpen: false,
```
ericauv, line 101: @stredger changing this back to `true` fixes the Invite Students ( `web/components/Students/Header` ) Popover issue.

However during this investigation, noticed that when this is changed back to true, Popover is constantly re-rendered because of change on `L151`, which removed `prevProps` check.

Next steps:
* I'll open a PR to change isOpen back to `true` by default
* I'll create a new ticket to investigate how to not have Popover re-rendering endlessly while it is open, while also still having it appear in the correct position on first render in Timeline/Milestone tooltips

PR #61, opened 2020-07-28
Add xsmall chip size

stredger:
## Description
Add xsmall chip size

### DIFF
--- a/packages/riipen-ui-docs/src/pages/components/chip/Compact.jsx
+++ b/packages/riipen-ui-docs/src/pages/components/chip/Compact.jsx
@@ -23,11 +23,13 @@ export default function Compact() {
         <Chip label="Tertiary" size="small" color="tertiary" />
       </div>
       <div style={breakStyle}>
-        <Chip label="Positive" size="small" color="positive" />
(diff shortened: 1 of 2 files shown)


packages/riipen-ui-docs/src/pages/components/chip/Compact.jsx
```diff
@@ -23,11 +23,13 @@ export default function Compact() {
         <Chip label="Tertiary" size="small" color="tertiary" />
       </div>
       <div style={breakStyle}>
-        <Chip label="Positive" size="small" color="positive" />
+        <Chip label="Positive" size="xsmall" color="positive" />
```
ericauv, line 26: Feel that this section would be more clear if we use the same color for all the chips, and only vary the size (like we do with disabled)

e.g.
```suggestion
return(
      <div>
      <Chip label="medium" size="medium" color="default"/>
      <span style={style}/>
      <Chip label="small" size="small" color="default"/>
      <span style={style}/>
      <Chip label="xsmall" size="xsmall" color="default"/>
      </div>
      );
```

stredger: 👍

PR #62, opened 2020-07-31
Allow tabs to be forced to display active

stredger:
## Notes
Using this for nested tabs where the values dont always match (such as using window.location for the value)

## Screenshots
Screen Shot 2020-07-31 at 11 10 13 AM

### DIFF
--- a/packages/riipen-ui/src/components/Tabs.jsx
+++ b/packages/riipen-ui/src/components/Tabs.jsx
@@ -86,8 +86,10 @@ class Tabs extends React.Component {
     const childrenWithProps = React.Children.map(children, child => {
       if (!child) return null;
 
+      const active = child.props.active || child.props.value === value;
+
       return React.cloneElement(child, {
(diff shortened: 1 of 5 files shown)


packages/riipen-ui/src/components/Tabs.jsx
```diff
@@ -86,8 +86,10 @@ class Tabs extends React.Component {
     const childrenWithProps = React.Children.map(children, child => {
       if (!child) return null;
 
+      const active = child.props.active || child.props.value === value;
```
ericauv, line 89: Feel like we should have some way of preventing multiple tabs in a list from being `active`, but not sure how to do that. Think we could do that in a separate ticket

stredger: So this worked the way you described before as active was always being overwritten. Maybe i should name the prop `forceActive` or something

PR #64, opened 2020-08-12
[PLT-5608] Editor Placeholder

jordanell:
## Description

1. Fixed issue with editor CSS where placeholder styles were not working
2. Updated doc pages
3. Moved the editor components to be flat instead of nested in file structure

## Notes

I am trying to stay as close to material-ui as possible with this repo, so moving to flat files structures seems right.


packages/riipen-ui/src/components/Editor.jsx
```diff
@@ -282,32 +282,35 @@ class Editor extends React.Component {
       }
 
       /* Rich Text Editor */
-      :global .DraftEditor-root {
+      :global(.DraftEditor-root) {
```
jordanell, line 285: These were the actual fixes. I think we may have upgraded `styled-jsx` at some point and they changes their interface to this for global styles.

PR #67, opened 2020-08-26
Import specific fontawesome icons instead of all

kradical:
https://github.com/riipen/developer-web/pull/71

## Description
This makes it so that only the icons used are imported instead of the whole library (700kb parsed) 🎉 .

## Notes
This will probably help out platform too because hopefully we are using the proper import style everywhere on that project.


kradical:
merging this, @bolinkd lmk if you see any issues and we can un-merge it!

bolinkd:
> merging this, @bolinkd lmk if you see any issues and we can un-merge it!

looks good to me, good catch!

PR #68, opened 2020-08-31
Input label color

stredger:
## Description
Add color to input

## Screenshots
Screen Shot 2020-08-31 at 3 51 14 PM

### DIFF
--- a/packages/riipen-ui/src/components/InputHint.jsx
+++ b/packages/riipen-ui/src/components/InputHint.jsx
@@ -41,8 +41,6 @@ class InputHint extends React.Component {
 
     const className = clsx(classes, color);
 
-    console.log(className);
-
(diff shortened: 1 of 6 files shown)


packages/riipen-ui/src/components/InputHint.jsx
```diff
@@ -45,6 +51,10 @@ class InputHint extends React.Component {
             color: ${theme.palette.text.secondary};
             margin-bottom: ${theme.spacing(2)}px;
           }
+
+          .white {
```
ksmort, line 55: Should there be a whitelist of available colours or nah?

stredger: ah yea probably in the proptypes

PR #72, opened 2020-09-25
Add span wrapper around img avatars

ksmort:
## Description
There is currently an issue with the UI kit avatar where if there is an avatar with an image and an avatar with just letters (or other content), they do not display nicely next to each other.  This PR attempts to solve this issue by updating the css in the Avatar component.

Original issue: 
Screenshot_2020-09-28 Avatar Riipen UI(2)

## Screenshots
Screenshot_2020-09-28 Avatar Riipen UI(1)

## Where to Start
packages/riipen-ui/src/components/Avatar.jsx

### DIFF
--- a/packages/riipen-ui/src/components/Avatar.jsx
+++ b/packages/riipen-ui/src/components/Avatar.jsx
@@ -71,24 +71,21 @@ class Avatar extends React.Component {
 
     return (
       <React.Fragment>
-        {src ? (
-          <span className="image">
+        <span className={className}>
+          {src ? (
             <img alt={alt} src={src} className={className} {...other} />
-          </span>
-        ) : (
-          <span className={className}>{children}</span>
-        )}
+          ) : (
+            <span className="inner">{children}</span>
+          )}
+        </span>
         <style jsx>{`
           .avatar {
-            align-items: center;
             background-color: ${theme.palette.grey[200]};
             border: ${this.getBorderSize()} solid ${theme.palette.common.white};
-            display: inline-flex;
-            flex-shrink: 0;
+            display: inline-block;
             font-family: ${theme.typography.fontFamily};
             font-size: calc(${size} * 0.4);
             height: ${size};
-            justify-content: center;
             line-height: 1;
             overflow: hidden;
             user-select: none;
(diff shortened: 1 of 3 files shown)


packages/riipen-ui/src/components/Avatar.jsx
```diff
@@ -96,6 +98,13 @@ class Avatar extends React.Component {
           .circle {
             border-radius: 50%;
           }
+
+          .image {
+            display: block;
+            height: calc(${size} + ${this.getBorderSize()} * 2);
+            width: calc(${size} + ${this.getBorderSize()} * 2);
```
ksmort, line 105: Have to take border into account.

jordanell: It makes me slightly nervous that this is `display: block`. I think as a user of the UI kit this would be unexpected. I am not entirely sure what this PR is really solving so I don't have any alternative suggestions.

@ksmort maybe can you elaborate a bit more on this?

packages/riipen-ui/src/components/Avatar.jsx
```diff
@@ -71,22 +71,21 @@ class Avatar extends React.Component {
 
     return (
       <React.Fragment>
-        {src ? (
-          <img alt={alt} src={src} className={className} {...other} />
-        ) : (
-          <span className={className}>{children}</span>
-        )}
+        <span className={className}>
```
ksmort, line 74: Throw base styling into a wrapper span

packages/riipen-ui/src/components/Avatar.jsx
```diff
@@ -71,22 +71,21 @@ class Avatar extends React.Component {
...
+          )}
+        </span>
         <style jsx>{`
           .avatar {
-            align-items: center;
             background-color: ${theme.palette.grey[200]};
             border: ${this.getBorderSize()} solid ${theme.palette.common.white};
-            display: inline-flex;
-            flex-shrink: 0;
+            display: inline-block;
```
ksmort, line 85: Use inline-block for wrapper span

packages/riipen-ui/src/components/Avatar.jsx
```diff
@@ -96,6 +95,21 @@ class Avatar extends React.Component {
...
           }
+
+          .image {
+            height: ${size};
+            width: ${size};
+          }
+
+          .inner {
+            align-items: center;
+            display: flex;
```
ksmort, line 106: Use flex on inner span (the non-image one)

packages/riipen-ui/src/components/Avatar.jsx
ksmort: @jordanell Updated PR description and updated css in Avatar component, hopefully it makes more sense!

jordanell (approved):

PR #73, opened 2020-09-29
Update avatar image height

ksmort:
## Description
Small update image height so image fits within wrapper span contents.

## Screenshots
Screenshot_2020-09-29 Avatar Riipen UI(1)

## Where to Start
packages/riipen-ui/src/components/Avatar.jsx

### DIFF
--- a/packages/riipen-ui/src/components/Avatar.jsx
+++ b/packages/riipen-ui/src/components/Avatar.jsx
@@ -97,8 +97,8 @@ class Avatar extends React.Component {
           }
 
           .image {
-            height: ${size};
-            width: ${size};
+            height: 100%;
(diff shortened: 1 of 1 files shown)


packages/riipen-ui/src/components/Avatar.jsx
```diff
@@ -97,8 +97,8 @@ class Avatar extends React.Component {
           }
 
           .image {
-            height: ${size};
-            width: ${size};
+            height: 100%;
+            width: 100%;
```
ksmort, line 101: The span css already sets the height, so just can put image height to 100%

jordanell (approved):

PR #74, opened 2020-09-30
Add min/max width/height to Avatar component

ksmort:
## Description
Hopefully last pr...

Update avatar component to set min/max widths/height so that avatar doesn't collapse in with margins/other styling (seen in staging currently in course cards).

## Screenshots
Prevents this:

### DIFF
--- a/packages/riipen-ui/src/components/Avatar.jsx
+++ b/packages/riipen-ui/src/components/Avatar.jsx
@@ -87,9 +87,12 @@ class Avatar extends React.Component {
             font-size: calc(${size} * 0.4);
             height: ${size};
             line-height: 1;
+            max-height: ${size};
+            max-width: ${size};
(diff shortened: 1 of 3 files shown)


packages/riipen-ui/src/components/Avatar.jsx
```diff
@@ -87,9 +87,12 @@ class Avatar extends React.Component {
             font-size: calc(${size} * 0.4);
             height: ${size};
             line-height: 1;
+            max-height: ${size};
+            max-width: ${size};
+            min-height: ${size};
+            min-width: ${size};
             overflow: hidden;
             user-select: none;
-            width: ${size};
```
ksmort, line 92: Need to keep height to make sure the letters are centered properly, but no longer need this width.

jordanell (approved):

PR #80, opened 2020-10-16
[PLT-5804] Refactor Form into function component

stredger:
## Description
1. Convert Form to function component

## Notes
This is pretty much all thats needed to scroll properly when the error actually gets passed into this component properly.
Validation errors do not however (as they happen sync on the fields themselves, while the input is being changed)
Debouncing doesnt really solve the issue (just delays it) so this might be "good enough" to solve the scrolling issue at least

### DIFF
--- a/packages/riipen-ui/src/components/Form.jsx
+++ b/packages/riipen-ui/src/components/Form.jsx
@@ -30,7 +30,6 @@ const Form = ({
     if ((error || !isEmpty(errors)) && errorScroll) {
       const padding = theme.spacing(20);
 
-      console.log(errorRef);
       const rect = errorRef.current.getBoundingClientRect();
 
       // Only scroll if it's not already visible
--- a/packages/riipen-ui/src/components/Form.jsx
+++ b/packages/riipen-ui/src/components/Form.jsx
@@ -133,19 +133,11 @@ const Form = ({
   }
 
   if (!isEmpty(errors)) {
-    if (
-      Array.isArray(errors) &&
-      errors.length > 0 &&
-      typeof errors[0] === "string"
-    ) {
+    if (Array.isArray(errors) && typeof errors[0] === "string") {
       errorsRenderer = renderErrorsString;
-    } else if (
-      Array.isArray(errors) &&
-      errors.length > 0 &&
-      React.isValidElement(errors[0])
-    ) {
+    } else if (Array.isArray(errors) && React.isValidElement(errors[0])) {
       errorsRenderer = renderErrorsNode;
-    } else if (typeof errors === "object" && Object.keys(errors).length > 0) {
+    } else if (typeof errors === "object") {
       errorsRenderer = renderErrorsObject;
     }
   }
--- a/packages/riipen-ui/src/components/Form.jsx
+++ b/packages/riipen-ui/src/components/Form.jsx
@@ -1,6 +1,6 @@
 import clsx from "clsx";
 import PropTypes from "prop-types";
-import React from "react";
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/Form.jsx
```diff
@@ -9,80 +9,37 @@ import withClasses from "../utils/withClasses";
...
-  componentDidUpdate(prevProps) {
-    const theme = this.context;
-
-    // If an error id dynamically added to the form, scroll the error notice in to view
-    if (
-      !prevProps.error &&
-      !prevProps.errors &&
-      (this.props.error || this.props.errors) &&
-      this.error &&
-      this.props.errorScroll
```
stredger, line 67: There was a bug here as props.errors is an object and therefor always true once it had been set. So !prevProps.errors was always false

packages/riipen-ui/src/components/Form.jsx
```diff
@@ -163,82 +116,117 @@ class Form extends React.Component {
...
+      errors.length > 0 &&
+      typeof errors[0] === "string"
+    ) {
+      errorsRenderer = renderErrorsString;
+    } else if (
+      Array.isArray(errors) &&
+      errors.length > 0 &&
+      React.isValidElement(errors[0])
+    ) {
+      errorsRenderer = renderErrorsNode;
```
ericauv, line 147: can remove errors.length>0 since `!Object.keys(['i have length']).length === false`, so it's checked already by `!isEmpty`
```suggestion
    if (
      Array.isArray(errors)
    ) {
      if(typeof errors[0] === "string"){
            errorsRenderer = renderErrorsString;
      } else if(React.isValidElement(errors[0])){
            errorsRenderer = renderErrorsNode;      
      }
```

packages/riipen-ui/src/components/Form.jsx
```diff
@@ -163,82 +116,117 @@ class Form extends React.Component {
...
+    } else if (typeof errors === "object" && Object.keys(errors).length > 0) {
```
ericauv, line 148: `isEmpty` checks this length, `typeof errors === "object"` filters out `undefined/false` value
```suggestion
    } else if (typeof errors === "object") {
```

packages/riipen-ui/src/components/Form.jsx
```diff
@@ -9,80 +9,37 @@ import withClasses from "../utils/withClasses";
...
-    if (
-      !prevProps.error &&
-      !prevProps.errors &&
-      (this.props.error || this.props.errors) &&
-      this.error &&
-      this.props.errorScroll
-    ) {
+const isEmpty = value => {
+  return !value || !Object.keys(value).length;
+};
```
ericauv, line 14: could use `lodash/isEmpty` here

packages/riipen-ui/src/components/Form.jsx
```diff
@@ -9,80 +9,37 @@ import withClasses from "../utils/withClasses";
...
-    // If an error id dynamically added to the form, scroll the error notice in to view
-    if (
-      !prevProps.error &&
-      !prevProps.errors &&
-      (this.props.error || this.props.errors) &&
-      this.error &&
-      this.props.errorScroll
-    ) {
+const isEmpty = value => {
+  return !value || !Object.keys(value).length;
```
ericauv, line 13:
```suggestion
  return !Object.keys(value).length;
```

ericauv (approved):
Looks good! Some redundant logic can be removed

PR #83, opened 2020-10-16
Add ContainerHeader component

ksmort:
## Description
Add ContainerHeader component that gives containers a coloured header.
Update Container component to have colour, border, and header options.
Convert Container to functional component.

## Screenshots
Screenshot_2020-10-16 Container Riipen UI(1)
Screenshot_2020-10-16 Container Riipen UI(2)
Screenshot_2020-10-16 Container Riipen UI(3)

## Where to Start
components/ContainerHeader
components/Container - last commit converts to functional component.

### DIFF
--- a/packages/riipen-ui/src/components/Container.jsx
+++ b/packages/riipen-ui/src/components/Container.jsx
@@ -77,7 +77,7 @@ Container.propTypes = {
   /**
    * The color of the container.
    */
-  color: PropTypes.oneOf("default", "white"),
+  color: PropTypes.oneOf(["default", "white"]),
 
   /**
    * The props to pass to the ContainerHeader.
--- a/packages/riipen-ui/src/components/Container.jsx
+++ b/packages/riipen-ui/src/components/Container.jsx
@@ -5,16 +5,7 @@ import React from "react";
 import ThemeContext from "../styles/ThemeContext";
 import withClasses from "../utils/withClasses";
 
-import ContainerHeader from "./ContainerHeader";
-
-const Container = ({
-  border,
-  classes,
-  children,
-  color,
-  headerProps,
-  maxWidth
-}) => {
+const Container = ({ border, classes, children, color, header, maxWidth }) => {
   const theme = React.useContext(ThemeContext);
 
   const className = clsx("root", classes);
@@ -22,7 +13,7 @@ const Container = ({
   return (
     <React.Fragment>
       <div className={className}>
-        {headerProps && <ContainerHeader {...headerProps} />}
+        {header}
         <div className={clsx(border && "border", color)}>{children}</div>
       </div>
       <style jsx>{`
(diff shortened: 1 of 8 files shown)


packages/riipen-ui/src/components/Container.jsx
```diff
@@ -5,70 +5,99 @@ import React from "react";
...
+}) => {
+  const theme = React.useContext(ThemeContext);
+
+  const className = clsx("root", classes);
+
+  return (
+    <React.Fragment>
+      <div className={className}>
+        {headerProps && <ContainerHeader {...headerProps} />}
+        <div className={clsx(border && "border", color)}>{children}</div>
```
ksmort, line 26: Wrap children in div to add border and color classes

packages/riipen-ui/src/components/Container.jsx
```diff
@@ -5,70 +5,99 @@ import React from "react";
...
+  maxWidth
+}) => {
+  const theme = React.useContext(ThemeContext);
+
+  const className = clsx("root", classes);
+
+  return (
+    <React.Fragment>
+      <div className={className}>
+        {headerProps && <ContainerHeader {...headerProps} />}
```
ksmort, line 25: conditionally render header

packages/riipen-ui/src/components/Container.jsx
```diff
@@ -5,70 +5,99 @@ import React from "react";
...
+        <div className={clsx(border && "border", color)}>{children}</div>
+      </div>
+      <style jsx>{`
+        .root {
```
ksmort, line 29: rename to root so it doesn't apply to all divs

packages/riipen-ui/src/components/Container.jsx
stredger: I wonder if this would work better if we just passed in the header. So something like 
```
<Container header={<ContainerHeader />} />
```
instead of passing all the props to the Container, then having that pass on to the Header

stredger (changes requested):
Cool. I suggestion about making the container more generic. Otherwise 💯

packages/riipen-ui/src/components/Container.jsx
stredger: For some context, i think we should do this with the input as well. We have been adding a lot of InputLabel props to control the style, where we should probably just pass the label and hint in so we dont have to keep messing with the parent component

ksmort: Yeah that sounds good!

packages/riipen-ui/src/components/Container.jsx
```diff
@@ -5,70 +5,90 @@ import React from "react";
...
-          }
+const Container = ({ border, classes, children, color, header, maxWidth }) => {
+  const theme = React.useContext(ThemeContext);
+
+  const className = clsx("root", classes);
+
+  return (
+    <React.Fragment>
+      <div className={className}>
+        {header}
```
ksmort, line 16: Pass in header prop instead of headerProps

stredger (approved):
🔥

PR #85, opened 2020-10-19
Update container border

ksmort:
## Description
Add more defined border to container when border prop is specified.

## Screenshots
Screenshot_2020-10-19 Container Riipen UI

## Where to Start
components/Container


stredger (approved):
Sweet 👍. Either fix the audit in a separate PR then rebase this bad boy, or just fix in here

PR #90, opened 2020-10-29
[PLT-5814] Add VerticalProgressBar to the UI-Kit

daelynj:
_[Jira Ticket: [PLT-5814]](https://riipen.atlassian.net/browse/PLT-5814)_

## Description
The `ProgressBarVertical` component implemented [here](https://github.com/riipen/platform-web/pull/2986) by @ksmort needed to be added to the UI-Kit.

## Notes
To be honest I floundered quite a bit troubleshooting the CSS and figuring out that some of it wasn't being picked up by the `Grid` component. I ended up using `linkedStyles` to make it happen once I got some assistance.

For the most part I tried to mirror other components and could use some context as to what this achieves:
```
export default withClasses(VerticalProgressBarItem);
```
I wrote pretty small commits in case I really need to change things in multiple places.

## Screenshots
Screen Shot 2020-10-29 at 3 06 37 PM

### DIFF
--- a/packages/riipen-ui/src/components/VerticalProgressBar.jsx
+++ b/packages/riipen-ui/src/components/VerticalProgressBar.jsx
@@ -1,0 +1,62 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import withClasses from "../utils/withClasses";
+
+import Grid from "./Grid";
+import GridItem from "./GridItem";
+import VerticalProgressBarItem from "./VerticalProgressBarItem";
+
+const VerticalProgressBar = ({ progresses }) => {
+  const renderProgress = () => {
+    return progresses.map((progress, index) => (
+      <GridItem key={`progress-${index}`}>
+        <VerticalProgressBarItem
+          progress={{ ...progress, noBar: index >= progresses.length - 1 }}
+        />
+      </GridItem>
+    ));
+  };
+  return (
+    <div>
+      <Grid spacing={0}>{renderProgress()}</Grid>
+    </div>
+  );
+};
+
+VerticalProgressBar.propTypes = {
+  // external
+
+  /**
+   * The progresses to display.
+   */
+  progresses: PropTypes.arrayOf(
+    PropTypes.shape({
+      /**
+       * The progress icon.
+       */
+      icon: PropTypes.object,
+
+      /**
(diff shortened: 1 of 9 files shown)


packages/riipen-ui/src/components/VerticalProgressBar.jsx
```diff
@@ -0,0 +1,62 @@
...
+    return progresses.map((progress, index) => (
+      <GridItem key={`progress-${index}`}>
+        <VerticalProgressBarItem
+          progress={{ ...progress, noBar: index >= progresses.length - 1 }}
+        />
+      </GridItem>
+    ));
+  };
+  return (
+    <div>
```
ksmort, line 21: Might be able to get rid of this exterior div?

packages/riipen-ui/src/components/VerticalProgressBarItem.jsx
```diff
@@ -0,0 +1,143 @@
...
+
+        .iconWrapper {
+          align-items: center;
+          border-radius: 50%;
+          display: flex;
+          height: 24px;
+          justify-content: center;
+          margin: ${theme.spacing(2)}px 0 ${theme.spacing(7)}px
+            ${theme.spacing(-4.65)}px;
+          padding: ${theme.spacing(1.5)}px;
```
ksmort, line 105: Usually we want to keep the spacing at integers, you can probably just use the number value of pixels instead of decimals or see if the styling is okay with the integer values

ksmort (approved):
Looks awesome! 🙌  Just two tiny things and then 🛥️

packages/riipen-ui/src/components/VerticalProgressBar.jsx
daelynj: Done 👍

packages/riipen-ui/src/components/VerticalProgressBarItem.jsx
daelynj: Sounds good - it's slightly not perfect then but hard to notice unless you sit an inch from your screen.

ksmort: Haha, woohoo!

PR #93, opened 2020-11-24
[PLT-6320] Added radio buttons and groups

bolinkd:
## Description
[PLT-6320](https://riipen.atlassian.net/browse/PLT-6320)

## Screenshots
Screen Shot 2020-11-24 at 09 17 58

## Where to Start
packages/riipen-ui/src/components/RadioButton.jsx

### DIFF
--- a/packages/riipen-ui-docs/src/pages/components/radio/Button.jsx
+++ b/packages/riipen-ui-docs/src/pages/components/radio/Button.jsx
@@ -6,16 +6,12 @@ import Form from "riipen-ui/components/Form";
 import RadioButton from "riipen-ui/components/RadioButton";
 import RadioButtonGroup from "riipen-ui/components/RadioButtonGroup";
 
-const icon = i => () => <FontAwesomeIcon icon={i} />;
+const icon = i => <FontAwesomeIcon icon={i} />;
 
 export default function Button() {
-  const [state, setState] = React.useState({
-    value: "none"
-  });
+  const [state, setState] = React.useState("none");
 
-  const handleChange = (e, value) => {
-    setState({ ...state, value });
-  };
+  const handleChange = (e, value) => setState(value);
 
   return (
     <Form>
--- a/packages/riipen-ui-docs/src/pages/components/radio/Button.jsx
+++ b/packages/riipen-ui-docs/src/pages/components/radio/Button.jsx
@@ -1,0 +1,54 @@
+import { faFlag } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import React from "react";
+
+import Form from "riipen-ui/components/Form";
+import RadioButton from "riipen-ui/components/RadioButton";
+import RadioButtonGroup from "riipen-ui/components/RadioButtonGroup";
+
+const icon = i => () => <FontAwesomeIcon icon={i} />;
+
+export default function Button() {
+  const [state, setState] = React.useState({
+    value: "none"
+  });
+
+  const handleChange = (e, value) => {
+    setState({ ...state, value });
+  };
+
+  return (
+    <Form>
+      <RadioButtonGroup
+        label="Variant"
+        hint="Pick your favorite variant."
+        onChange={handleChange}
+        required
+      >
+        <RadioButton
+          prefix={icon(faFlag)}
+          label="Prefix"
+          value="prefix"
+          checked={state.value === "prefix"}
+        />
+        <RadioButton
+          suffix={icon(faFlag)}
+          label="Suffix"
+          value="suffix"
+          checked={state.value === "suffix"}
+        />
+        <RadioButton
+          label="None"
+          value="none"
+          checked={state.value === "none"}
+        />
+        <RadioButton
+          disabled
+          label="Disabled"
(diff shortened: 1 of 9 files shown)


packages/riipen-ui-docs/src/pages/components/radio/Button.jsx
```diff
@@ -0,0 +1,54 @@
...
+const icon = i => () => <FontAwesomeIcon icon={i} />;
+
+export default function Button() {
+  const [state, setState] = React.useState({
+    value: "none"
+  });
+
+  const handleChange = (e, value) => {
+    setState({ ...state, value });
+  };
```
ericauv, line 18:
```suggestion
  const [value, setValue] = React.useState('none');

  const handleChange = (e, value) => {
    setValue(value);
  };
```

ericauv: + change all `state.value` --> `value`

packages/riipen-ui/src/components/RadioButton.jsx
```diff
@@ -0,0 +1,135 @@
...
+      className={clsx(
+        classes,
+        "root",
+        checked && "checked",
+        disabled && "disabled"
+      )}
+    >
+      <label
+        className={clsx(checked && "checked", disabled && "disabled")}
+        htmlFor={other.id}
```
ericauv, line 34: since we're doing `other.id`, I think we should add `id` to propTypes + destructure it out of props, and just use `id`

packages/riipen-ui/src/components/RadioButton.jsx
```diff
@@ -0,0 +1,135 @@
...
+  return (
+    <div
+      className={clsx(
+        classes,
+        "root",
+        checked && "checked",
+        disabled && "disabled"
+      )}
+    >
+      <label
```
ericauv, line 32: + does the label need to wrap everything?

packages/riipen-ui/src/components/RadioButtonGroup.jsx
```diff
@@ -0,0 +1,128 @@
...
+      onChange: handleChange,
+      classes: [linkedStyles.className, "radioButton"]
+    })
+  );
+
+  return (
+    <React.Fragment>
+      <fieldset>
+        {(label || hint) && (
+          <InputLabel hint={hint} htmlFor={other.name} required={required}>
```
ericauv, line 55: same comment re: `other.propName` --> `propName`

ericauv (approved):
Looks good to me, small comments:
* destructuring props out of `props`, and adding to `propTypes` wherever we do `other.propName`
* question about `label` wrapping everything

packages/riipen-ui/src/components/RadioButton.jsx
```diff
@@ -0,0 +1,135 @@
...
+
+  /**
+   * The component to render before the label text.
+   */
+  prefix: PropTypes.elementType,
+
+  /**
+   * The props to supply to the prefix component.
+   */
+  prefixProps: PropTypes.object,
```
jordanell, line 120: Instead of this, should we use inversion of control with `prefix: PropTypes.node`? Let the user render their prefix at the same level at the `<RadioButton>`, then we can avoid this extra prop passing.

packages/riipen-ui/src/components/RadioButton.jsx
```diff
@@ -0,0 +1,135 @@
...
+
+  /**
+   * The component to render after the label text.
+   */
+  suffix: PropTypes.elementType,
+
+  /**
+   * The props to supply to the suffix component.
+   */
+  suffixProps: PropTypes.object
```
jordanell, line 130: Same here with inversion of control.

packages/riipen-ui/src/components/RadioButtonGroup.jsx
```diff
@@ -0,0 +1,128 @@
...
+  label,
+  onChange,
+  required,
+  warning,
+  ...other
+}) => {
+  const linkedStyles = css.resolve`
+    .radioButtons .radioButton:first-child {
+      border-top-left-radius: 15px;
+      border-bottom-left-radius: 15px;
```
jordanell, line 25: I wounder if we should put these border radius sizes into our theme? `shape.borderRadius.xxl`? Your call. It's sort of a one off.

jordanell (changes requested):

packages/riipen-ui/src/components/RadioButton.jsx
bolinkd: I use the label instead of wrapping everything in a div (I could change it but then it would be exactly the same except all the styles applied to the div instead of the label with the label also inside the div. I think this is the cleanest way to do it

packages/riipen-ui/src/components/RadioButtonGroup.jsx
bolinkd: there is no standardized set of radii so I am going to opt to leaving it out for now, if we get a set to use then I am all for this change

bolinkd: I dont think we have very many rounded edges in the UI Kit at the moment

PR #98, opened 2020-12-04
Update editor styles to match new designs

bolinkd:
## Description
design:
Screen Shot 2020-12-04 at 13 59 08

Main changes:
No border - use colour difference instead
Ability to add left justified actions
Ability to add control row full height actions

## Notes
Have to update a couple places the editor is used on the web app where the prop naming has changed, but other than that its a minor change

## Screenshots
Screen Shot 2020-12-04 at 13 51 58

### DIFF
--- a/packages/riipen-ui/src/components/Editor.jsx
+++ b/packages/riipen-ui/src/components/Editor.jsx
@@ -398,7 +398,7 @@ class Editor extends React.Component {
     }
   };
 
-  forceFocus = () => focus(true);
+  forceFocus = () => this.focus(true);
 
   // Control button callback for toggling block type
   toggleBlockType = blockType => {
--- a/packages/riipen-ui/src/components/Editor.jsx
+++ b/packages/riipen-ui/src/components/Editor.jsx
@@ -96,9 +96,19 @@ class Editor extends React.Component {
 
   static propTypes = {
     /**
-     * Additional controls to display.
+     * Whether or not to move focus the Editor on first render.
      */
-    additionalControls: PropTypes.node,
+    autoFocus: PropTypes.bool,
+
+    /**
+     * action controls to display on the right side of the editor control bar.
+     */
+    actionControls: PropTypes.arrayOf(PropTypes.node),
+
+    /**
+     * Additional action controls to display on the left side of the editor control bar.
+     */
+    stylingControls: PropTypes.arrayOf(PropTypes.node),
 
     /**
      * The id of the element to use as the aria-label for the Editor.
@@ -149,7 +159,9 @@ class Editor extends React.Component {
   };
 
   static defaultProps = {
-    additionalControls: [],
+    autoFocus: true,
+    stylingControls: [],
+    actionControls: [],
     controlPosition: "top"
   };
 
@@ -186,11 +198,12 @@ class Editor extends React.Component {
 
   getLinkedStyles = () => {
     const theme = this.context;
+    const { controlPosition } = this.props;
 
     return css.resolve`
       .wrapper {
         background-color: ${theme.palette.common.white};
-        border: 1px solid ${theme.palette.grey[400]};
+        border: 1px solid #979797;
         border-radius: ${theme.shape.borderRadius.md};
         font-family: ${theme.typography.body1.fontFamily};
         font-size: ${theme.typography.body1.fontSize};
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/Editor.jsx
```diff
@@ -186,11 +198,12 @@ class Editor extends React.Component {
 
   getLinkedStyles = () => {
     const theme = this.context;
+    const { controlPosition } = this.props;
 
     return css.resolve`
       .wrapper {
         background-color: ${theme.palette.common.white};
-        border: 1px solid ${theme.palette.grey[400]};
+        border: 1px solid #979797;
```
bolinkd, line 206: doesnt exist anywhere in the UI Kit 🙈

packages/riipen-ui/src/components/Editor.jsx
```diff
@@ -242,17 +259,20 @@ class Editor extends React.Component {
...
 
-      .controlContainer > div {
+      .stylingControls > div:not(:last-child) {
         border-right: 1px solid ${theme.palette.divider};
       }
 
-      .controlContainer > div:last-child {
-        border-right: none;
-      }
-
```
bolinkd, line 255: same, but different

packages/riipen-ui/src/components/Editor.jsx
```diff
@@ -96,9 +96,14 @@ class Editor extends React.Component {
...
   static propTypes = {
     /**
-     * Additional controls to display.
+     * Whether or not to move focus the Editor on first render.
      */
-    additionalControls: PropTypes.node,
+    autoFocus: PropTypes.bool,
+
+    /**
+     * action controls to display on the right side of the editor control bar.
```
jordanell, line 104: Capitol "A" for "Action".

packages/riipen-ui/src/components/Editor.jsx
```diff
@@ -145,12 +150,19 @@ class Editor extends React.Component {
...
+    /**
+     * Additional action controls to display on the left side of the editor control bar.
+     */
+    stylingControls: PropTypes.arrayOf(PropTypes.node)
   };
 
   static defaultProps = {
-    additionalControls: [],
-    controlPosition: "top"
+    autoFocus: true,
```
jordanell, line 162: Should this default to true? If the editor is being displayed in a form and it is say the 3rd form element, I don't think we should auto focus it.

packages/riipen-ui/src/components/Editor.jsx
jordanell: Ya this is slightly annoying. Is there any color that even comes close? I would rather use a "close" color that is in the kit than a random hard coded color.

packages/riipen-ui/src/components/Editor.jsx
```diff
@@ -242,17 +259,20 @@ class Editor extends React.Component {
...
       .controlContainer {
-        margin: 0 ${theme.spacing(2)}px;
+        background-color: ${theme.palette.grey[100]};
+        display: flex;
+        flex: 1 1 auto;
+        flex-direction: row;
+        justify-content: space-between;
+        padding-left: ${theme.spacing(2)}px;
+        border-bottom-left-radius: 4px;
+        border-bottom-right-radius: 4px;
```
jordanell, line 269: Use `${theme.shape.borderRadius.md}` here instead.

jordanell (changes requested):

packages/riipen-ui/src/components/Editor.jsx
bolinkd: its the current behaviour, I have no problem changing it, but this was mainly to prevent it from occurring on the messages page as it was stealing focus from the search input when the url changed

packages/riipen-ui/src/components/Editor.jsx
```diff
@@ -186,11 +198,12 @@ class Editor extends React.Component {
 
   getLinkedStyles = () => {
     const theme = this.context;
+    const { controlPosition } = this.props;
```
ksmort, line 201: Should updating this component to be a function component be added as a tech debt ticket?

packages/riipen-ui/src/components/Editor.jsx
ksmort: https://i.ytimg.com/vi/IGXYzNm29bQ/hqdefault.jpg

packages/riipen-ui/src/components/Editor.jsx
```diff
@@ -615,7 +649,7 @@ class Editor extends React.Component {
 
     return (
       <React.Fragment>
-        <div className={wrapperClasses} onClick={this.focus}>
+        <div className={wrapperClasses} onClick={() => this.focus(true)}>
```
ksmort, line 652: _Possibly_ make this not anonymous?

ksmort (changes requested):
Overall looks good!  Basically agree with everything Jordan pointed out.

packages/riipen-ui/src/components/Editor.jsx
jordanell: Ya I think default to false, then we change the message page to use `autoFocus={true}`.

packages/riipen-ui/src/components/Editor.jsx
bolinkd: I believe it is a good idea, but we should have a general epic for updating each of the components to function components

bolinkd: I will add it to the 'sprint retro' list

jordanell (approved):

ksmort (approved):

PR #99, opened 2020-12-07
Update hover functionality in Chip component

ksmort:
## Description
Add section to Chip page for `onClick` prop.
Update the icon section to use `icon` prop instead of `iconStart`.
Added in the hover support.

## Notes
Not 100% on what colour we want the hover to be, maybe checkout the branch and do a look?
Also a TODO: convert to function component.

## Screenshots
Screenshot_2020-12-07 Chip Riipen UI

## Where to Start
91 packages/riipen-ui/src/components/Chip.jsx

### DIFF
--- a/packages/riipen-ui-docs/src/pages/components/chip/Hover.jsx
+++ b/packages/riipen-ui-docs/src/pages/components/chip/Hover.jsx
@@ -28,6 +28,26 @@ export default function Hover() {
         <Chip label="Warning" hover color="warning" />
         <span style={style} />
         <Chip label="Negative" hover color="negative" />
+        <span style={style} />
+        <Chip label="Dark" hover color="dark" />
+      </div>
+      <div style={breakStyle}>
+        <Chip label="Default" hover variant="outlined" />
+        <span style={style} />
+        <Chip label="Primary" hover color="primary" variant="outlined" />
+        <span style={style} />
+        <Chip label="Secondary" hover color="secondary" variant="outlined" />
+        <span style={style} />
+        <Chip label="Tertiary" hover color="tertiary" variant="outlined" />
+      </div>
+      <div style={breakStyle}>
+        <Chip label="Positive" hover color="positive" variant="outlined" />
+        <span style={style} />
+        <Chip label="Warning" hover color="warning" variant="outlined" />
+        <span style={style} />
+        <Chip label="Negative" hover color="negative" variant="outlined" />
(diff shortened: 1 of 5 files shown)


packages/riipen-ui-docs/src/pages/components/chip/Hover.jsx
```diff
@@ -28,6 +28,26 @@ export default function Hover() {
         <Chip label="Warning" hover color="warning" />
         <span style={style} />
         <Chip label="Negative" hover color="negative" />
+        <span style={style} />
```
ksmort, line 31: This file already existed, so I'm going to assume hover support was supposed to exist, but maybe got forgotten. :)

packages/riipen-ui/src/components/Chip.jsx
```diff
@@ -198,85 +209,137 @@ class Chip extends React.Component {
             border-color: ${theme.palette.primary.main};
             color: ${theme.palette.primary.contrast};
           }
+          .primary.hover:hover {
```
ksmort, line 212: Not sure if we want to darken chips on hover, thoughts?

packages/riipen-ui/src/components/Chip.jsx
```diff
@@ -198,85 +209,137 @@ class Chip extends React.Component {
...
+            background-color: ${theme.palette.positive.dark};
+            border-color: ${theme.palette.positive.dark};
+          }
           .positive.outlined {
             border-color: ${theme.palette.positive.main};
             color: ${theme.palette.positive.main};
           }
-
+          .positive.outlined.hover:hover {
+            background-color: transparent;
```
ksmort, line 312: Also not sure on colour change for outlined version.

packages/riipen-ui/src/components/Chip.jsx
bolinkd: I was thinking that _outlined_ chips would become _default_ and _default_ would become _outlined_. 

But that also depends on what the designs specify, I would ask Maribel to decide what she wants to do in this situation

ksmort: I shall ask her, convert this PR to draft for now, and then update when ready!

ksmort: Updated to be the inverted variant!

bolinkd (approved):

PR #101, opened 2020-12-17
[PLT-6474] Editor: mobileControlRow prop

ericauv:
## Description
* Add prop to not `display: none` controlRow when on mobile
* web -> Message/Form/New shouldn't hide the control row when on mobile https://github.com/riipen/platform-web/pull/3393
## Screenshots
Screen Shot 2020-12-17 at 11 34 46 AM

### DIFF
--- a/packages/riipen-ui/src/components/Editor.jsx
+++ b/packages/riipen-ui/src/components/Editor.jsx
@@ -590,8 +590,6 @@ class Editor extends React.Component {
       ? ["controlContainer", "mobileControlRow"]
       : "controlContainer";
 
-    console.log({ props: this.props });
-
     return (
       <React.Fragment>
         <div className={clsx(linkedStyles.className, controlContainerClasses)}>
--- a/packages/riipen-ui/src/components/Editor.jsx
+++ b/packages/riipen-ui/src/components/Editor.jsx
@@ -132,6 +132,11 @@ class Editor extends React.Component {
     error: PropTypes.any,
 
     /**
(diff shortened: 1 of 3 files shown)


packages/riipen-ui/src/components/Editor.jsx
```diff
@@ -569,14 +576,25 @@ class Editor extends React.Component {
...
+
     const { editorState } = this.state;
 
     const linkedStyles = this.getLinkedStyles();
 
+    const controlContainerClasses = hideControlRowOnMobile
+      ? ["controlContainer", "hideControlRowOnMobile"]
+      : "controlContainer";
+
+    console.log({ props: this.props });
```
bolinkd, line 593: I think jordan usually uses `console.log('wtf', props)`

packages/riipen-ui/src/components/Editor.jsx
```diff
@@ -131,6 +131,11 @@ class Editor extends React.Component {
      */
     error: PropTypes.any,
 
+    /**
+     * Whether or not to hide the controls + control row when on smaller screens.
+     */
+    hideControlRowOnMobile: PropTypes.bool,
```
bolinkd, line 137: I would go with mobileControlRow which defaults to false

packages/riipen-ui/src/components/Editor.jsx
ericauv: 🙈

bolinkd (approved):

PR #105, opened 2021-01-19
test chip component with enzyme

licarm:
## Description
Set up Enzyme and Jest with riipen-ui.
Test Chip component

## Notes
Fix "import/no-extraneous-dependencies": "off" in .eslintrc.json

## Where to Start
chip.test.jsx


packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,125 @@
...
+    expect(
+      wrapper
+        .find("Chip")
+        .find("span")
+        .text()
+    ).toEqual(label);
+    wrapper.unmount();
+  });
+
+  it("displays a label given as a child of Chip", () => {
```
jordanell, line 52: Again for naming convention, you could do:


```js
describe('label prop', () => {
  it('should display label from child', () => {
    ...
  });
});
```

Try to apply this philosophy to all remaining tests in this PR.

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,125 @@
...
+        .find("Chip")
+        .childAt(0)
+        .hasClass(custom)
+    ).toEqual(true);
+    wrapper.unmount();
+  });
+
+  it("gives an error when setting it to an invalid color", () => {
+    const errors = jest.spyOn(console, "error").mockImplementation();
+    shallow(<Chip color="red" />);
```
jordanell, line 86: `mount` instead of `shallow`. Apply as needed throughout PR.

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,125 @@
...
+    expect(
+      wrapper
+        .find("Chip")
+        .childAt(0)
+        .hasClass("outlined")
+    ).toEqual(true);
+  });
+
+  it("component can be any element", () => {
+    const wrapper = mount(<Chip component="a"> </Chip>);
```
jordanell, line 101: This should be 

```js
const wrapper = mount(<Chip component="a" />);
```

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,125 @@
...
+      wrapper
+        .find("Chip")
+        .childAt(0)
+        .name()
+    ).toEqual("a");
+  });
+});
+
+describe("Chip component is interactive", () => {
+  it("triggers click handler on click", () => {
```
jordanell, line 114: This might be worth looking at for conventions on how to name these sorts of tests.

If I was to continue using my suggestion above, I might instead do:

```js
describe('onClick prop', () => {
  it('should invoke onClick handler when clicked', () => {
   ...
  });
});
```

I'm happy with any convention we pick as long as we stick with it :).

jordanell (changes requested):
I left a bunch of comments on how we can clean this PR up.

There are also a bunch of small config things that I will just go in an do to help this out as they are definitely more advanced and I don't want us to get slowed down by some config stuff.

Once you are done making updates to the PR, re-request my review and I can take another look!

packages/riipen-ui/src/components/Chip.test.jsx
imogenem: Adjusted to your suggestion

jordanell:
Closing in favor of https://github.com/riipen/ui/pull/106.

PR #106, opened 2021-01-19
Jest Infrastructure

jordanell:
## Description

1. Setup monorepo for jest testing


packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,136 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import Chip from "./Chip";
+
```
jordanell, line 6: For all the tests, I would actually be tempted to put them all inside a `describe` block sort of like you had before:

```js
describe('<Chip>', () => {
  // all tests / it blocks in here
})
```

According to the jest docs it is good to wrap objects (in our case components) in these blocks for organization https://jestjs.io/docs/en/api#describename-fn.

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+it("renders without errors", () => {
+  try {
+    mount(<Chip />);
+  } catch (error) {
+    expect(true).toEqual(false);
```
jordanell, line 11: This won't actually work. If the chip does not throw an error, the `catch` block never actually triggers.

I think what you want is:

```js
let error;

try {
  mount(<Chip />);
} catch (e) {
  error = e;
}

expect(error).toEqual(undefined);
```

In the example above `expect(error).toEqual(undefined);` always runs and the error actually only gets set if the `catch` block triggers.

You could also do this same thing with a `try / catch / finally` as described here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch but that is probably overkill.

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+  }
+});
+
+describe("default props", () => {
+  it("sets correct default props", () => {
+    const defaultProps = new Chip().type.defaultProps;
+    defaultProps.classes.push("riipen-chip", "riipen");
```
jordanell, line 18: Does this actually do anything? Just looking at this test it's hard to tell exactly what this is doing. I am probably just stupid though.

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+    expect(component.hasClass(defaultProps.size)).toEqual(true);
+    expect(component.hasClass(defaultProps.variant)).toEqual(true);
+    expect(component.childAt(0).name()).toEqual("span");
+
+    expect(toJson(wrapper)).toMatchSnapshot();
+  });
+});
+
+describe("variant prop", () => {
+  it("outlines chip given outlined variant", () => {
```
jordanell, line 36: I am thinking once we fully flesh out the tests for the `<Chip>` component, we will probably want to do something similar as we did with the `color` prop below:

1. Test a setting for the prop
2. Test an out of bounds condition

In the case of 1, I would be tempted to rename this to `"applies variant class name"` or something more generic that isn't specific to outlines. That way our tests look like:

1. Does it work given a good value
2. Does it work given a bad value

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+      wrapper
+        .find("Chip")
+        .childAt(0)
+        .hasClass("outlined")
+    ).toEqual(true);
+  });
+});
+
+describe("component prop", () => {
+  it("component can be any element", () => {
```
jordanell, line 49: For the name of this test, again I would try to be as consistant as possible:

1. Does it work given a good value
2. Does it work given a bad value

So I would maybe try to align this with the wording for the color and variant prop as previously described. Maybe something like `"uses custom component"`.

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+    expect(component.hasClass(defaultProps.variant)).toEqual(true);
+    expect(component.childAt(0).name()).toEqual("span");
+
+    expect(toJson(wrapper)).toMatchSnapshot();
+  });
+});
+
+describe("variant prop", () => {
+  it("outlines chip given outlined variant", () => {
+    const wrapper = mount(<Chip variant="outlined" />);
```
jordanell, line 37: I likw what you did down below with color:

```js
const custom = "dark";

const wrapper = mount(<Chip color={custom} />);
```

For the sake of consistancy, I would recommend you do something similar here:

```js
const custom = "outlined";

const wrapper = mount(<Chip variant={custom} />);
```

Then in the assertion below you can use the `custom` variable.

See my comments below in the color tests regarding the variable name of `custom`.

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+        .find("Chip")
+        .childAt(0)
+        .name()
+    ).toEqual("a");
+  });
+});
+
+describe("color prop", () => {
+  it("sets custom color", () => {
+    const custom = "dark";
```
jordanell, line 64: `custom` is sort of a generic variable name. I would get in to the habit of being more verbose with your variable names. I would maybe instead call this `color` instead of `custom` as without even looking at the value I know that it will be a color.

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+      wrapper
+        .find("Chip")
+        .childAt(0)
+        .name()
+    ).toEqual("a");
+  });
+});
+
+describe("color prop", () => {
+  it("sets custom color", () => {
```
jordanell, line 63: Seeing as your test below for the inverse of this is `"gives an error when given an invalid color"`, I would maybe think about renaming this to `"sets color class name with a valid color"`.

If we have two tests which are inverses of each other, the naming should reflect that.

I would apply this to all of my previous comments up above.

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+    mount(<Chip color="red" />);
+
+    expect(errors).toHaveBeenCalledTimes(1);
+  });
+});
+
+describe("label prop", () => {
+  it("displays label given as the label prop", () => {
+    const label = "A test label";
```
jordanell, line 88: Nice variable name :).

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+    const wrapper = mount(<Chip label={label} />);
+
+    expect(wrapper.find("Chip").props().label).toEqual(label);
```
jordanell, line 92: This isn't really testing anything. It's effectively saying "I pass a prop to a component, now make sure I passed a prop to a component".

I think we can remove this.

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+        .children()
+        .contains(child)
+    ).toEqual(true);
+    expect(
+      wrapper
+        .find(".label")
+        .childAt(0)
+        .text()
+    ).toEqual(label);
+  });
```
jordanell, line 119: I don't think this test is quite right.

Looking at the source code for `<Chip>`, it states:

```js
const label = this.props.label || this.props.children;
```

Meaning that:

1. if a label is provided, it uses that
2. if children are provided, it uses that
3. if label and children are provided it uses the label

So I think our tests should probably reflect that.

We should actually change this test to `"prioritizes displaying label over children"` where was pass both label and children into the props and make sure only label shows up.

Then, we should actually make another `describe('children prop' () => {...});` which tests `"displays child"` or something.

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+});
+
+describe("onClick event", () => {
```
jordanell, line 122: I'm not sure this is correct. I think we are actually testing `describe('onClick prop', ...)` so as to be consistent with other describe blocks.

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+  it("invokes onClick handler when clicked", () => {
+    const handler = jest.fn();
+
+    const wrapper = mount(<Chip onClick={handler} />);
+
+    expect(handler).toHaveBeenCalledTimes(0);
```
jordanell, line 128: I don't think we actually need this. Not sure it's adding value.

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+    wrapper.find("Chip").simulate("click");
+    expect(handler).toHaveBeenCalledTimes(1);
+
+    wrapper.find("Chip").simulate("click");
+    expect(handler).toHaveBeenCalledTimes(2);
```
jordanell, line 134: Why test twice? Seems unnecessary. But open for debate.

jordanell (commented):
Made another pass at the testing logic for review purposes. 

You should read all comments before addressing any as lots of them are inter-related.

packages/riipen-ui/src/components/Chip.test.jsx
imogenem: removed test

PR #107, opened 2021-01-21
Remove onIconClick in Chip component

licarm:
## Description
Removed onIconClick from Chip component because its not used anymore


jordanell:
@licarm something got messed up here with git. See how this PR has a commit ("Setup monorepo for jest testing ") that isn't part of what you just did?

Not sure what happened here, but we can't merge this one as is because there's all sorts of weird git-ness happening.

We have a few options:

1. Close this PR, re-branch off master, and re-do the work you did
2. Get real fancy and re-branch off master, then git cherry-pick the commits you want over (I can help you with this)

Let me know which you prefer :).

PR #108, opened 2021-01-21
remove onIconClick prop from Chip

licarm:
## Description
Removed onIconClick from Chip component because its not used anymore

### DIFF
--- a/packages/riipen-ui/src/components/Chip.test.jsx
+++ b/packages/riipen-ui/src/components/Chip.test.jsx
@@ -5,10 +5,6 @@ import React from "react";
 import Chip from "./Chip";
 
 describe("<Chip>", () => {
-  afterEach(() => {
-    jest.restoreAllMocks();
-  });
-
   it("renders without errors", () => {
     let error;
     try {
--- a/packages/riipen-ui/src/components/Chip.test.jsx
+++ b/packages/riipen-ui/src/components/Chip.test.jsx
@@ -5,6 +5,10 @@ import React from "react";
 import Chip from "./Chip";
 
 describe("<Chip>", () => {
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
   it("renders without errors", () => {
     let error;
     try {
@@ -12,6 +16,7 @@ describe("<Chip>", () => {
     } catch (e) {
       error = e;
     }
+
     expect(error).toEqual(undefined);
   });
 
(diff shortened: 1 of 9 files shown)


jordanell:
@licarm this looks great!

We should also remove any wording in the documentation site around `onIconClick`.

See here https://ui.riipen.com/components/chip/#icon ? Maybe we can change it to just read "You can provide an icon at the start of the Chip.".

The code for this will be in `riipen-ui-docs`.

Remember:

1. You can run the docs site locally by doing `npm start` from the root of the repository
2. I would do a project wide search for the text "onIconClick" which will locate the file you need to update.

Good luck!

jordanell (changes requested):
@licarm this looks great!

We should also remove any wording in the documentation site around onIconClick.

See here https://ui.riipen.com/components/chip/#icon ? Maybe we can change it to just read "You can provide an icon at the start of the Chip.".

The code for this will be in riipen-ui-docs.

Remember:

    You can run the docs site locally by doing npm start from the root of the repository
    I would do a project wide search for the text "onIconClick" which will locate the file you need to update.

Good luck!

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -5,13 +5,18 @@ import React from "react";
 import Chip from "./Chip";
 
 describe("<Chip>", () => {
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
```
jordanell, line 10: We will have to include this same block of code inside of every single test file we write, which isn't very DRY.

Perhaps there is a better way we can do this.

Take a look at this SoF question and the accepted answer: https://stackoverflow.com/questions/48722930/running-a-global-test-setup-before-each-test-in-jest

I think we should be able to add: 

```js
global.afterEach(() => {
   jest.restoreAllMocks();
});
```

inside of our `test/setup.js` file. That way we only have to write this code once.

packages/riipen-ui/src/components/Chip.test.jsx
```diff
@@ -34,23 +39,86 @@ describe("<Chip>", () => {
...
+    });
+
+    it("gives an error when classes are provided as one string", () => {
+      const classes = "classOne classTwo";
+      const errors = jest.spyOn(console, "error").mockImplementation();
+
+      mount(<Chip classes={classes} />);
+
+      expect(errors).toHaveBeenCalled();
+    });
```
jordanell, line 91: Niiiiiice!

jordanell (changes requested):
This is awesome! Just one suggestion to keep our tests DRY in the future.

jordanell (approved):

PR #110, opened 2021-01-22
[PLT-6624] Unit Tests for AppBar 

imogenem:
## Description

Added in unit tests for rendering of AppBar and all default and variant props.

## Notes

Especially interested to get feedback on the default prop tests.

## Where to Start

AppBar.test.jsx in riipen-ui


packages/riipen-ui/src/components/AppBar.test.jsx
```diff
@@ -0,0 +1,92 @@
...
+
+    const component = wrapper.find("AppBar").childAt(0);
+    expect(component.hasClass(defaultProps.color)).toEqual(true);
+    expect(component.hasClass(defaultProps.position)).toEqual(true);
+
+    expect(toJson(wrapper)).toMatchSnapshot();
+  });
+
+  describe("children prop", () => {
+    it("displays custom label from child", () => {
```
jordanell, line 31: This name is sort of confusing. Looking at the details for `<AppBar>` if you pass it `children` it simply renders them. There is no mention of a "label".

Would maybe rename this to "displays given children" or something.

packages/riipen-ui/src/components/AppBar.test.jsx
```diff
@@ -0,0 +1,92 @@
...
+      const label = "A label given as a child";
+      const child = <div>{label}</div>;
```
jordanell, line 33: I think you can actually just combine these. 

```js
const child = <div>A child</div>;

const wrapper = mount(<AppBar>{child}</AppBar>);

expect(wrapper.containsMatchingElement(child)).toBeTruthy();
```

since we aren't testing this specific label thing.

packages/riipen-ui/src/components/AppBar.test.jsx
```diff
@@ -0,0 +1,92 @@
...
+  describe("color prop", () => {
+    it("sets color class name with valid custom colour", () => {
+      const colorVariant = "secondary";
+
+      const wrapper = mount(<AppBar color={colorVariant} />);
+
+      expect(wrapper.find("AppBar").props().color).toEqual(colorVariant);
+    });
+
+    it("sets color class name with invalid colour", () => {
```
jordanell, line 66: Should this be "throws an error with invalid color" or something? This test name doesn't match what the test is actually doing.

packages/riipen-ui/src/components/AppBar.test.jsx
```diff
@@ -0,0 +1,92 @@
...
+    });
+  });
+
+  describe("position prop", () => {
+    it("sets position class name with valid custom position", () => {
+      const positionVariant = "absolute";
+
+      const wrapper = mount(<AppBar color={positionVariant} />);
+
+      expect(wrapper.find("AppBar").props().color).toEqual(positionVariant);
```
jordanell, line 81: I don't think this is right? You should be testing `position` here but are testing `color`. Propbably just a copy paste error :).

jordanell (changes requested):

packages/riipen-ui/src/components/AppBar.test.jsx
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  describe("color prop", () => {
+    it("sets color class name with valid custom colour", () => {
+      const colorVariant = "secondary";
+
+      const wrapper = mount(<AppBar color={colorVariant} />);
+
+      expect(wrapper.find("AppBar").props().color).toEqual(colorVariant);
+    });
```
jordanell, line 62: I'm not sure this is entirely right. The test name says it's testing for a CSS class name, but the test itself is just making sure a prop is set. I think the actual test should be looking at the CSS class name.

packages/riipen-ui/src/components/AppBar.test.jsx
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  describe("position prop", () => {
+    it("sets position class name with valid custom position", () => {
+      const positionVariant = "absolute";
+
+      const wrapper = mount(<AppBar position={positionVariant} />);
+
+      expect(wrapper.find("AppBar").props().position).toEqual(positionVariant);
+    });
```
jordanell, line 80: Same thing here. Test name and test don't match. Should be actually testing the CSS class name.

jordanell (changes requested):

jest.config.js
```diff
@@ -1,5 +1,6 @@
 module.exports = {
-  moduleDirectories: ['node_modules', './'],
-  modulePaths: ['node_modules', './'],
-  projects: [ '<rootDir>/packages/*/jest.config.js']
-};
+  displayName: "riipen-ui",
+  name: "riipen-ui",
+  roots: ["<rootDir>/packages/riipen-ui"],
+  setupFilesAfterEnv: ["<rootDir>/packages/riipen-ui/test/setup.js"]
+};
```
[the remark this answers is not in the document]
imogenem: Noted and adjusted

packages/riipen-ui/src/components/AppBar.test.jsx
imogenem: Adjusted to follow how Chip tested this.

jordanell (approved):

PR #114, opened 2021-01-25
[PLT-6626] Unit tests for Backdrop

licarm:
## Description
Unit tests for Backdrop

## Where to Start
Backdrop.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/Backdrop.test.jsx
+++ b/packages/riipen-ui/src/components/Backdrop.test.jsx
@@ -76,7 +76,8 @@ describe("<Backdrop>", () => {
 
       const wrapper = mount(<Backdrop open={open} />);
 
-      expect(wrapper.find("Backdrop").props().open).toEqual(true);
+      expect(toJson(wrapper)).toMatchSnapshot();
+      expect(wrapper.find("Backdrop").hasClass("open")).toEqual(true);
     });
   });
 });
--- a/packages/riipen-ui/src/components/Backdrop.test.jsx
+++ b/packages/riipen-ui/src/components/Backdrop.test.jsx
@@ -16,13 +16,10 @@ describe("<Backdrop>", () => {
     expect(error).toEqual(undefined);
   });
 
-  describe("default props", () => {
-    it("sets correct default props", () => {
-      const wrapper = mount(<Backdrop />);
+  it("snapshot test", () => {
+    const wrapper = mount(<Backdrop />);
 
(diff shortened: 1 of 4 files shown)


packages/riipen-ui/src/components/Backdrop.test.jsx
```diff
@@ -0,0 +1,82 @@
...
+
+  describe("open prop", () => {
+    it("sets open prop to correct value", () => {
+      const open = true;
+
+      const wrapper = mount(<Backdrop open={open} />);
+
+      expect(wrapper.find("Backdrop").props().open).toEqual(true);
+    });
+  });
```
jordanell, line 81: I think we could do a better job with this test. The `open` prop actually gets applied to the root node's `className` via the `clsx` package. You might want to read through that package's documetation, but basically if you pass it a `boolean`, it will render the variable's name into the `className`. So in our case, if `open = true`, the `className` of the root node will `hasClass("open")`. Does that make sense?

We should also add the `false` test to make sure when `open = false` that the root node does not get the `open` class applied to it.

jordanell (changes requested):
Overall looking good. Added comment around one of the tests to make it a bit stronger.

packages/riipen-ui/src/components/Backdrop.test.jsx
licarm: when I look at the snapshot for `mount(<mount(<Backdrop open={true} />)`, I get this:
```
<withClasses(Backdrop)
  open={true}
>
  <Backdrop
    classes={
      Array [
        "riipen",
        "riipen-backdrop",
      ]
    }
    open={true}
  >
    <div
      aria-hidden={true}
      className="jsx-1867827232 root riipen riipen-backdrop"
    />
    <JSXStyle
      id="1867827232"
    />
  </Backdrop>
</withClasses(Backdrop)>
```
`wrapper.find("Backdrop").hasClass('open')` also gives me a false. Is it possible Backdrop is not applying the open prop properly?

jordanell: Oh woah, this is actually a sort of "bug". Good catch here. Now that I look closer, in the `<style jsx>` element, there actually isn't a `.open` class! This means we probably had it at one point, but remove it.

We should just fully remove the `open` prop as it actually doesn't do anything!

Great catch.

jordanell (approved):
Nice job!

PR #116, opened 2021-01-26
[PLT-6625] Unit tests for Avatar

imogenem:
## Description
Unit tests for Avatar Component

## Notes
Looking for feedback on alt and src prop tests

## Where to Start
Avatar.test.jsx in components folder

### DIFF
--- a/packages/riipen-ui/src/components/Avatar.test.jsx
+++ b/packages/riipen-ui/src/components/Avatar.test.jsx
@@ -1,0 +1,167 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import Avatar from "./Avatar";
+
+describe("<Avatar>", () => {
+  it("renders without errors", () => {
+    let error;
+    try {
+      mount(<Avatar />);
+    } catch (e) {
+      error = e;
+    }
+    expect(error).toEqual(undefined);
+  });
+
+  describe("default props", () => {
+    it("sets correct default props", () => {
+      const wrapper = mount(<Avatar />);
+
+      expect(wrapper.find("Avatar").props().size).toEqual("96px");
+      expect(wrapper.find("Avatar").props().variant).toEqual("circle");
+
+      expect(toJson(wrapper)).toMatchSnapshot();
+    });
+
+    it("appends higher order values to default classes prop with withClass decorator", () => {
+      const wrapper = mount(<Avatar />);
+
+      expect(wrapper.find("Avatar").props().classes).toEqual([
+        "riipen",
+        "riipen-avatar"
+      ]);
+    });
+  });
+
+  describe("alt prop", () => {
+    it("sets inner class name given custom alt props", () => {
+      const altVariant = "Test String";
+
+      const wrapper = mount(<Avatar alt={altVariant} />);
+
+      expect(
+        wrapper
+          .find(".avatar")
+          .childAt(0)
+          .hasClass("inner")
+      ).toEqual(true);
+      expect(wrapper.find("Avatar").props().alt).toEqual(altVariant);
+    });
+  });
+
+  describe("children prop", () => {
+    it("displays given children", () => {
+      const initials = "ZZ";
+      const child = <div>{initials}</div>;
+
+      const wrapper = mount(<Avatar>{child}</Avatar>);
+
+      expect(
+        wrapper
+          .find(".inner")
+          .children()
+          .contains(child)
+      ).toEqual(true);
+      expect(
+        wrapper
+          .find(".inner")
+          .childAt(0)
+          .text()
+      ).toEqual(initials);
+    });
+  });
+
+  describe("classes prop", () => {
+    it("sets custom class", () => {
+      const classVariant = ["newClass"];
+
+      const wrapper = mount(<Avatar classes={classVariant} />);
+
+      expect(
+        wrapper
+          .find("Avatar")
+          .childAt(0)
+          .hasClass(classVariant[0])
+      ).toEqual(true);
+    });
+  });
+
+  describe("size prop", () => {
+    it("sets valid custom size", () => {
+      const sizeVariant = "50px";
+
+      const wrapper = mount(<Avatar size={sizeVariant} />);
+
+      expect(wrapper.find("Avatar").props().size).toEqual(sizeVariant);
+    });
+
+    it("sets valid border size above 70", () => {
+      const sizeVariant = "75px";
+
+      const wrapper = mount(<Avatar size={sizeVariant} />);
+
+      expect(
+        wrapper
+          .find("Avatar")
+          .instance()
+          .getBorderSize()
+      ).toEqual("4px");
+    });
+
+    it("sets valid border size below 70", () => {
+      const sizeVariant = "65px";
+
+      const wrapper = mount(<Avatar size={sizeVariant} />);
+
+      expect(
+        wrapper
+          .find("Avatar")
+          .instance()
+          .getBorderSize()
+      ).toEqual("2px");
+    });
+  });
+
+  describe("src prop", () => {
+    it("sets image class name given custom alt and src props", () => {
+      const altVariant = "Test String";
+      const srcVariant = "/static/images/avatars/jordan-ell.png";
+
+      const wrapper = mount(<Avatar alt={altVariant} src={srcVariant} />);
+
+      expect(
+        wrapper
+          .find(".avatar")
+          .childAt(0)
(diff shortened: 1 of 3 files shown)


packages/riipen-ui/src/components/Avatar.test.jsx
```diff
@@ -0,0 +1,167 @@
...
+    });
+
+    it("appends higher order values to default classes prop with withClass decorator", () => {
+      const wrapper = mount(<Avatar />);
+
+      expect(wrapper.find("Avatar").props().classes).toEqual([
+        "riipen",
+        "riipen-avatar"
+      ]);
+    });
```
jordanell, line 35: I like this. We should do this on every component.

packages/riipen-ui/src/components/Avatar.test.jsx
```diff
@@ -0,0 +1,167 @@
...
+
+      expect(
+        wrapper
+          .find(".avatar")
+          .childAt(0)
+          .hasClass("inner")
+      ).toEqual(true);
+      expect(wrapper.find("Avatar").props().alt).toEqual(altVariant);
+    });
+  });
```
jordanell, line 52: I don't think this is fully correct. To me looking at the source code, the `alt` prop is only useful when `src` is also passed. If you pass both `alt` and `src`, then the `` tag has it's `alt` value set to the value of the prop` alt`.

So i _think_ what we need is a test that looks like:

```js
const wrapper = mount(<Avatar alt={altVariant} src={...} />);

// expect wrapper.find('img').props.alt to equal alt
```

Does that make sense?

packages/riipen-ui/src/components/Avatar.test.jsx
```diff
@@ -0,0 +1,167 @@
...
+
+      const wrapper = mount(<Avatar alt={altVariant} />);
+
+      expect(
+        wrapper
+          .find(".avatar")
+          .childAt(0)
+          .hasClass("inner")
+      ).toEqual(true);
+      expect(wrapper.find("Avatar").props().alt).toEqual(altVariant);
```
jordanell, line 50: I don't think we need this. This is that weird test again that is saying "I pass 'sum' the first arugment of "1", now tests did I pass 'sum' the first argument of 1".

packages/riipen-ui/src/components/Avatar.test.jsx
```diff
@@ -0,0 +1,167 @@
...
+    });
+  });
+
+  describe("children prop", () => {
+    it("displays given children", () => {
+      const initials = "ZZ";
+      const child = <div>{initials}</div>;
```
jordanell, line 57: I don't think you need to split these like this. I think you can just do `child = <div>ZZ</div>` then test to make sure that the wrapper contains that div.

packages/riipen-ui/src/components/Avatar.test.jsx
```diff
@@ -0,0 +1,167 @@
...
+          .find(".inner")
+          .children()
+          .contains(child)
+      ).toEqual(true);
+      expect(
+        wrapper
+          .find(".inner")
+          .childAt(0)
+          .text()
+      ).toEqual(initials);
```
jordanell, line 72: If you simplify the above, you can remove this test. This test is just a subset of the one above it, so we are actually testing the same thing twice.

packages/riipen-ui/src/components/Avatar.test.jsx
```diff
@@ -0,0 +1,167 @@
...
+  });
+
+  describe("size prop", () => {
+    it("sets valid custom size", () => {
+      const sizeVariant = "50px";
+
+      const wrapper = mount(<Avatar size={sizeVariant} />);
+
+      expect(wrapper.find("Avatar").props().size).toEqual(sizeVariant);
+    });
```
jordanell, line 98: This is that weird test again. Can remove.

packages/riipen-ui/src/components/Avatar.test.jsx
```diff
@@ -0,0 +1,167 @@
...
+
+    it("sets valid border size above 70", () => {
```
jordanell, line 100: Maybe "sets correct border size when size is above 70" ? Just slightly clearer.

packages/riipen-ui/src/components/Avatar.test.jsx
```diff
@@ -0,0 +1,167 @@
...
+
+      expect(
+        wrapper
+          .find("Avatar")
+          .instance()
+          .getBorderSize()
+      ).toEqual("4px");
+    });
+
+    it("sets valid border size below 70", () => {
```
jordanell, line 113: Maybe "sets correct border size when size is below 70" ? Just slightly clearer.

packages/riipen-ui/src/components/Avatar.test.jsx
```diff
@@ -0,0 +1,167 @@
...
+          .instance()
+          .getBorderSize()
+      ).toEqual("2px");
+    });
+  });
+
+  describe("src prop", () => {
+    it("sets image class name given custom alt and src props", () => {
+      const altVariant = "Test String";
+      const srcVariant = "/static/images/avatars/jordan-ell.png";
```
jordanell, line 130: lol oh gawd this image

packages/riipen-ui/src/components/Avatar.test.jsx
```diff
@@ -0,0 +1,167 @@
...
+    });
+  });
+
+  describe("variant prop", () => {
+    it("sets variant class name with valid custom variant", () => {
+      const variant = "rounded";
+
+      const wrapper = mount(<Avatar variant={variant} />);
+
+      expect(wrapper.find("Avatar").props().variant).toEqual(variant);
```
jordanell, line 150: This is that weird test again, can remove.

packages/riipen-ui/src/components/Avatar.test.jsx
```diff
@@ -0,0 +1,167 @@
...
+        wrapper
+          .find("Avatar")
+          .instance()
+          .getBorderSize()
+      ).toEqual("2px");
+    });
+  });
+
+  describe("src prop", () => {
+    it("sets image class name given custom alt and src props", () => {
```
jordanell, line 128: Nice this is what I was trying to explain up above with the `alt` tests.

I think we don't need the `alt` here if we make the changes I proposed above. This tests should simply be "renders an img tag". See test below for how to update.

packages/riipen-ui/src/components/Avatar.test.jsx
```diff
@@ -0,0 +1,167 @@
...
+
+      const wrapper = mount(<Avatar alt={altVariant} src={srcVariant} />);
+
+      expect(
+        wrapper
+          .find(".avatar")
+          .childAt(0)
+          .hasClass("image")
+      ).toEqual(true);
+      expect(wrapper.find("Avatar").props().src).toEqual(srcVariant);
```
jordanell, line 140: This is that weird test again, can remove.

packages/riipen-ui/src/components/Avatar.test.jsx
```diff
@@ -0,0 +1,167 @@
...
+      const srcVariant = "/static/images/avatars/jordan-ell.png";
+
+      const wrapper = mount(<Avatar alt={altVariant} src={srcVariant} />);
+
+      expect(
+        wrapper
+          .find(".avatar")
+          .childAt(0)
+          .hasClass("image")
+      ).toEqual(true);
```
jordanell, line 139: This should instead be "expect wrapper.avatar to contain a "img" tag and that the img tag has its `src` prop set to match the prop we gave".

jordanell (changes requested):

packages/riipen-ui/src/components/Avatar.test.jsx
imogenem: :}

packages/riipen-ui/src/components/Avatar.test.jsx
imogenem: Adjusted and renamed the test

packages/riipen-ui/src/components/Avatar.test.jsx
imogenem: expect(wrapper.find("Avatar").contains("img")) won't work in this case, it cannot find the img element.

As an alternative, can we use this? Is it necessary?
 expect(wrapper.find("Avatar").find("img")).toHaveLength(1);

jordanell (approved):
Great job!

PR #117, opened 2021-01-27
[PLT-6638] tests for Drawer component

licarm:
## Where to Start
Drawer.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/Drawer.jsx
+++ b/packages/riipen-ui/src/components/Drawer.jsx
@@ -1,6 +1,7 @@
 import clsx from "clsx";
 import PropTypes from "prop-types";
 import React from "react";
+import _JSXStyle from "styled-jsx/style";
 
 import ThemeContext from "../styles/ThemeContext";
 import withClasses from "../utils/withClasses";
--- a/packages/riipen-ui/src/components/Drawer.jsx
+++ b/packages/riipen-ui/src/components/Drawer.jsx
@@ -33,7 +33,7 @@ class Drawer extends React.Component {
      *
      * @param {object} event The event source of the callback.
      */
-    onClose: PropTypes.func,
+    onClose: PropTypes.func.isRequired,
 
     /**
      * If `true`, the drawer is open.
--- a/packages/riipen-ui/src/components/Drawer.test.jsx
+++ b/packages/riipen-ui/src/components/Drawer.test.jsx
@@ -6,9 +6,10 @@ import Drawer from "./Drawer.jsx";
 
 describe("<Drawer>", () => {
   it("renders without errors", () => {
+    const onClose = jest.fn();
     let error;
     try {
-      mount(<Drawer />);
+      mount(<Drawer onClose={onClose} />);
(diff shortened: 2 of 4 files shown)


packages/riipen-ui/src/components/Drawer.test.jsx
```diff
@@ -0,0 +1,177 @@
...
+    try {
+      mount(<Drawer />);
+    } catch (e) {
+      error = e;
+    }
+
+    expect(error).toEqual(undefined);
+  });
+
+  it("snapshot test when open is true and onClose is provided", () => {
```
jordanell, line 19: "it snapshot test when open is true and onClose is provided" is kind of a weird sentence to read.

Maybe we can do "it renders the correct snapshot when ...".

packages/riipen-ui/src/components/Drawer.test.jsx
```diff
@@ -0,0 +1,177 @@
...
+    const open = true;
+    const onClose = jest.fn();
```
jordanell, line 21: I just realized `onClose` is required for `<ClickAway>` but not in `Drawer`. I think we have 2 options here:

1. Make `onClose` a required prop for `<Drawer>`
2. Only render the `<ClickAway>` if `onClose` is provided, else just render the children.

Both have pros and cons. I think number 1 makes more sense to me.

packages/riipen-ui/src/components/Drawer.test.jsx
```diff
@@ -0,0 +1,177 @@
...
+
+      const wrapper = mount(<Drawer />);
+
+      expect(
+        wrapper
+          .find("Drawer")
+          .props()
+          .classes.sort()
+      ).toEqual(classes.sort());
+    });
```
jordanell, line 52: I'm not this this test should be nested within the `default props` describe blocks. We aren't really testing the default props here, we are testing something else.

packages/riipen-ui/src/components/Drawer.test.jsx
```diff
@@ -0,0 +1,177 @@
...
+      expect(wrapper.find("ClickAway").props().onClickAway).toEqual(onClose);
+    });
+
+    it("invokes onClose handler when component requests to be closed", () => {
+      const open = true;
+      const onClose = jest.fn();
+
+      const wrapper = mount(<Drawer open={open} onClose={onClose} />);
+
+      wrapper.find("ClickAway").simulate("click");
```
jordanell, line 150: Nice job!

jordanell (changes requested):

packages/riipen-ui/src/components/Drawer.jsx
```diff
@@ -32,7 +33,7 @@ class Drawer extends React.Component {
      *
      * @param {object} event The event source of the callback.
      */
-    onClose: PropTypes.func,
+    onClose: PropTypes.func.isRequired,
```
jordanell, line 36: Nice!

jordanell (approved):

PR #118, opened 2021-01-27
[PLT-6637] tests for Divider component

licarm:
## Where to Start
Divider.tests.jsx

### DIFF
--- a/packages/riipen-ui/src/components/Divider.test.jsx
+++ b/packages/riipen-ui/src/components/Divider.test.jsx
@@ -1,0 +1,98 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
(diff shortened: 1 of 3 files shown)


packages/riipen-ui/src/components/Divider.test.jsx
```diff
@@ -0,0 +1,98 @@
...
+
+      const wrapper = mount(<Divider />);
+
+      expect(
+        wrapper
+          .find("Divider")
+          .props()
+          .classes.sort()
+      ).toEqual(classes.sort());
+    });
```
jordanell, line 42: Not sure this should be bested inside the `default props` describe block. We aren't really testing the default props here.

jordanell (changes requested):

jordanell (approved):

PR #119, opened 2021-01-28
[PLT-6627] Unit Tests for Badge

imogenem:
## Description
Unit tests for Badge 

## Notes
Found a spelling error in the docs markdown file 

## Where to Start
Badge.test.jsx in the riipen-ui components folder
badge.md in the riipen-ui-docs components folder

### DIFF
--- a/packages/riipen-ui/src/components/Badge.test.jsx
+++ b/packages/riipen-ui/src/components/Badge.test.jsx
@@ -148,7 +148,7 @@ describe("<Badge>", () => {
   });
 
   describe("content prop", () => {
-    it("badge renders when content is 0 and showZero is false", () => {
+    it("hides badge when content is 0 and showZero is false", () => {
       const content = 0;
 
       const wrapper = mount(<Badge content={content} />);
@@ -160,10 +160,9 @@ describe("<Badge>", () => {
           .childAt(0)
           .hasClass("hidden")
       ).toEqual(true);
-      expect(Number(wrapper.text())).toEqual(content);
     });
 
-    it("badge renders when content is 0 and showZero is true", () => {
+    it("renders badge when content is 0 and showZero is true", () => {
       const content = 0;
       const showZero = true;
 
@@ -172,7 +171,7 @@ describe("<Badge>", () => {
       expect(Number(wrapper.find("Badge").text())).toEqual(content);
     });
 
-    it("badge renders when content is greater than 0 and showZero is true", () => {
+    it("renders badge when content is greater than 0 and showZero is true", () => {
       const content = 1;
       const showZero = true;
 
--- a/packages/riipen-ui/src/components/Badge.test.jsx
+++ b/packages/riipen-ui/src/components/Badge.test.jsx
@@ -148,8 +148,10 @@ describe("<Badge>", () => {
   });
 
   describe("content prop", () => {
-    it("badge does not render when content is null and showZero is false", () => {
-      const wrapper = mount(<Badge />);
+    it("badge renders when content is 0 and showZero is false", () => {
+      const content = 0;
+
+      const wrapper = mount(<Badge content={content} />);
 
       expect(
         wrapper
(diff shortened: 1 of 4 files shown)


packages/riipen-ui/src/components/Badge.test.jsx
```diff
@@ -0,0 +1,306 @@
...
+      const component = "abcdefg";
+
+      mount(<Badge component={component} />);
+
+      expect(errors).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe("content prop", () => {
+    it("badge does not render when content is null and showZero is false", () => {
```
jordanell, line 151: I think a better test would be "when content is 0". `null` will never render, but 0 will render if `showZero` is `true`.

packages/riipen-ui/src/components/Badge.test.jsx
```diff
@@ -0,0 +1,306 @@
...
+
+    it("badge renders with content and showZero props", () => {
+      const content = 2;
+      const showZero = true;
+
+      const wrapper = mount(<Badge content={content} showZero={showZero} />);
+
+      expect(Number(wrapper.find("Badge").text())).toEqual(2);
+    });
+  });
```
jordanell, line 172: I think we should add one more test here. "It should render truthy content" where content is just set to 1 or 2 or something. You covered the other two edge cases very well, but I think we need this base case as well.

packages/riipen-ui/src/components/Badge.test.jsx
```diff
@@ -0,0 +1,306 @@
...
+
+  describe("max prop", () => {
+    it("displays max value given max greater than content value", () => {
```
jordanell, line 175: Should this say "displays max value given content greater than max value"?

packages/riipen-ui/src/components/Badge.test.jsx
```diff
@@ -0,0 +1,306 @@
...
+      const max = 9;
+      const content = 11;
+
+      const wrapper = mount(<Badge max={max} content={content} />);
+
+      expect(wrapper.text()).toEqual(`${max}+`);
+    });
+
+    it("displays content value given content greater than max value", () => {
```
jordanell, line 184: I think this should say "display content value given content less than max value".

packages/riipen-ui/src/components/Badge.test.jsx
```diff
@@ -0,0 +1,306 @@
...
+      expect(
+        wrapper
+          .find("Badge")
+          .childAt(0)
+          .childAt(0)
+          .hasClass(size)
+      ).toEqual(true);
+    });
+
+    it("does not set size class name when invalid variant prop used", () => {
```
jordanell, line 251: Instead of "invalid variant" I would be tempted to say "non standard". This makes it more clear to anyone reading the test that invalid actually means non standard.

packages/riipen-ui/src/components/Badge.test.jsx
```diff
@@ -0,0 +1,306 @@
...
+          .childAt(0)
+          .hasClass(variant)
+      ).toEqual(true);
+      expect(
+        wrapper
+          .find("Badge")
+          .childAt(0)
+          .childAt(0)
+          .hasClass("size")
+      ).toEqual(false);
```
jordanell, line 295: I think we can remove this as we already tested it on line 257 above.

jordanell (changes requested):
Really nice job! Mostly just small things to clean up.

packages/riipen-ui/src/components/Badge.test.jsx
imogenem: Can I revise it to "non standard variant" to convey that it's the variant prop controlling if size can be used?

jordanell: Love it

packages/riipen-ui/src/components/Badge.test.jsx
imogenem: Added the the content prop and adjusted the test name a bit more.

packages/riipen-ui/src/components/Badge.test.jsx
```diff
@@ -0,0 +1,310 @@
...
+
+      expect(
+        wrapper
+          .find("Badge")
+          .childAt(0)
+          .childAt(0)
+          .hasClass("hidden")
+      ).toEqual(true);
+      expect(Number(wrapper.text())).toEqual(content);
+    });
```
jordanell, line 164: Slightly confused. According to the component, the "0" should not render if showZero is `false` (which is the default`. so why is line 163 actually working? Maybe it's because it actually _does_ render, but the CSS is hiding it.

I would maybe recommend just deleting line 163 as it's sort of confusing and rename the test maybe to "it hides the content when content is 0 and showZero is false".

Does that make sense?

jordanell (changes requested):
One last little thing.

packages/riipen-ui/src/components/Badge.test.jsx
imogenem: Yes it does

jordanell (approved):

PR #120, opened 2021-01-29
[PLT-6628] Banner component tests and inline style

licarm:
## Description
- tests for the Banner component
- height and src props are put into inline styles, instead of in the css

## Where to Start
Banner.jsx
Banner.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/Banner.jsx
+++ b/packages/riipen-ui/src/components/Banner.jsx
@@ -49,18 +49,23 @@ class Banner extends React.Component {
 
     const theme = this.context;
 
+    const style = {
+      height,
+      backgroundImage: `url(${src})`
+    };
+
     const className = clsx("banner", tint, classes);
 
     return (
       <React.Fragment>
-        <div className={className}>{children}</div>
+        <div className={className} style={style}>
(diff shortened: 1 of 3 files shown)


packages/riipen-ui/src/components/Banner.jsx
```diff
@@ -48,18 +49,23 @@ class Banner extends React.Component {
 
     const theme = this.context;
 
+    const style = {
+      height,
+      backgroundImage: `url(${src})`
+    };
```
jordanell, line 55: Slick!

packages/riipen-ui/src/components/Banner.test.jsx
```diff
@@ -0,0 +1,147 @@
...
+    it("sets the div inside Banner to a given height", () => {
+      const height = "50px";
+
+      const wrapper = mount(<Banner height={height} />);
+
+      expect(
+        wrapper
+          .find("Banner")
+          .childAt(0)
+          .props().style.height
```
jordanell, line 94: Nice!

jordanell (approved):
Nice job! This is a perfect test file :)

PR #121, opened 2021-01-29
[plt-6629] tests for Breadcrumbs

licarm:
## Description
tests for Breadcrumbs

## Where to Start
Breadcrumbs.test.jsx


packages/riipen-ui/src/components/Breadcrumbs.test.jsx
```diff
@@ -0,0 +1,189 @@
...
+      const component = wrapper.find("Breadcrumbs");
+      expect(component.props().component).toEqual(defaultProps.component);
+      expect(component.props().separator).toEqual(defaultProps.separator);
+
+      expect(toJson(wrapper)).toMatchSnapshot();
+    });
+  });
+
+  describe("children prop", () => {
+    it("renders children nodes as typography", () => {
```
jordanell, line 36: I think this test is actually testing that the children are rendered as a <li> not as <Typography>. We should up this title accordingly.

packages/riipen-ui/src/components/Breadcrumbs.test.jsx
```diff
@@ -0,0 +1,189 @@
...
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import Breadcrumbs from "./Breadcrumbs.jsx";
+
+describe("<Breadcrumbs>", () => {
+  it("renders without errors", () => {
+    let error;
+    const child = <h1>hello hello</h1>;
+    try {
```
jordanell, line 11: Following you conventions down below, it seems like you are always doing

```js
[arrange]
_blank line_
[act]
_blank line_
[assert]
```

Following that pattern, I think there should be a blank line between 10 and 11 as that's the gap between your arrange and assert here.

jordanell (changes requested):
Looks good! Two minor suggestions.

jordanell (approved):

PR #122, opened 2021-01-30
[PLT-6636] Unit Tests for Container

imogenem:
## Description
Added unit tests and added style attribute for the Container component.

## Where to Start
Container.jsx in components folder

### DIFF
--- a/packages/riipen-ui/src/components/Container.jsx
+++ b/packages/riipen-ui/src/components/Container.jsx
@@ -9,16 +9,19 @@ import withClasses from "../utils/withClasses";
 const Container = ({ classes, children, maxWidth }) => {
   const theme = React.useContext(ThemeContext);
 
+  const style = {
+    maxWidth: `${maxWidth ? `${theme.breakpoints[maxWidth]}px` : "100%"}`
+  };
+
   const className = clsx("root", classes);
 
   return (
-    <div className={className}>
+    <div className={className} style={style}>
       {children}
       <style jsx>{`
         .root {
           box-sizing: border-box;
           margin: 0 auto;
-          max-width: ${maxWidth ? `${theme.breakpoints[maxWidth]}px` : "100%"};
           padding: 0 ${theme.spacing(6)}px;
           position: relative;
         }
--- a/packages/riipen-ui/src/components/Container.jsx
+++ b/packages/riipen-ui/src/components/Container.jsx
@@ -1,6 +1,7 @@
 import clsx from "clsx";
 import PropTypes from "prop-types";
 import React from "react";
+import _JSXStyle from "styled-jsx/style";
 
 import ThemeContext from "../styles/ThemeContext";
 import withClasses from "../utils/withClasses";
--- a/packages/riipen-ui/src/components/Container.test.jsx
+++ b/packages/riipen-ui/src/components/Container.test.jsx
@@ -1,5 +1,8 @@
 import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
 import React from "react";
+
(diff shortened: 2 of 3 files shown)


packages/riipen-ui/src/components/Container.jsx
```diff
@@ -1,23 +1,27 @@
...
 import ThemeContext from "../styles/ThemeContext";
 import withClasses from "../utils/withClasses";
 
 const Container = ({ classes, children, maxWidth }) => {
   const theme = React.useContext(ThemeContext);
 
+  const style = {
+    maxWidth: `${maxWidth ? `${theme.breakpoints[maxWidth]}px` : "100%"}`
+  };
+
```
jordanell, line 15: Nice job

packages/riipen-ui/src/components/Container.test.jsx
```diff
@@ -0,0 +1,104 @@
...
+import React from "react";
+
+import defaultTheme from "../themes/default";
+
+import Container from "./Container";
+
+describe("<Container>", () => {
+  it("renders without errors", () => {
+    let error;
+    try {
```
jordanell, line 12: I think we should add a blank line between these two lines as this is the separating between our arrange and act steps of the test.

packages/riipen-ui/src/components/Container.test.jsx
```diff
@@ -0,0 +1,104 @@
...
+    });
+
+    it("appends higher order values to default classes prop with withClass decorator", () => {
+      const wrapper = mount(<Container />);
+
+      expect(wrapper.find("Container").props().classes).toEqual([
+        "riipen",
+        "riipen-container"
+      ]);
+    });
```
jordanell, line 40: @imogenem and you are writing these tests in slightly different places. Imo is putting this inside the `describe('classes prop'` and you are putting them in the `default props` block. We should be consistent. 

My two cents is probably to prefer it in the `classes` block since it's actually concatenating the HOC's classes to the component.

packages/riipen-ui/src/components/Container.test.jsx
```diff
@@ -0,0 +1,104 @@
...
+      const classes = ["newClass"];
+
+      const wrapper = mount(<Container classes={classes} />);
+
+      expect(
+        wrapper
+          .find("Container")
+          .childAt(0)
+          .hasClass(classes[0])
+      ).toEqual(true);
```
jordanell, line 64: You and @imogenem are also doing these differently. Imo is testing 2 classes and you are testing one. We should be consistent.

I think I prefer testing one rather than two as testing two doesn't add a ton of extra value.

I guess this comment is more for @imogenem on future PRs.

jordanell (changes requested):
NIce!

packages/riipen-ui/src/components/Container.test.jsx
imogenem: @licarm let's meet?

packages/riipen-ui/src/components/Container.test.jsx
imogenem: Adjusted

packages/riipen-ui/src/components/Container.test.jsx
imogenem: @jordanell Do you want previous test files adjusted to just test with one class?

packages/riipen-ui/src/components/Container.test.jsx
imogenem: Moved this test to the classes prop describe block.

jordanell (approved):

PR #123, opened 2021-02-02
[plt-6651] unit tests for MenuItem

licarm:
## Description
- tests for MenuItem component
- changed the way color prop is used in ListItem to use the default color when an invalid value is passed in

## Where to Start
- ListItem.jsx on line 67
- MenuItem.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/ListItem.jsx
+++ b/packages/riipen-ui/src/components/ListItem.jsx
@@ -64,7 +64,8 @@ const ListItem = props => {
         }
 
         .list-item.focusVisible {
-          color: ${theme.palette[color]?.main} || ${theme.palette.primary.main};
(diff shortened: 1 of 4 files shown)


packages/riipen-ui/src/components/ListItem.jsx
```diff
@@ -64,7 +64,8 @@ const ListItem = props => {
         }
 
         .list-item.focusVisible {
-          color: ${theme.palette[color].main};
+          color: ${theme.palette[color]?.main} ||
+            ${theme.palette[ListItem.defaultProps.color].main};
```
jordanell, line 68: Great job!

jordanell (approved):
Wowwzers. Getting pretty good at testing :)

PR #124, opened 2021-02-02
[plt-6669] tests for Toolbar component

licarm:
## Description
tests for Toolbar component

## Where to Start
Toolbar.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/Toolbar.test.jsx
+++ b/packages/riipen-ui/src/components/Toolbar.test.jsx
@@ -1,0 +1,77 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import Toolbar from "./Toolbar.jsx";
+
+describe("<Toolbar>", () => {
+  it("renders without errors", () => {
+    let error;
+
+    try {
+      mount(<Toolbar />);
+    } catch (e) {
+      error = e;
+    }
+    expect(error).toEqual(undefined);
+  });
+
+  it("renders correct snapshot", () => {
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/Toolbar.test.jsx
```diff
@@ -0,0 +1,77 @@
...
+      error = e;
+    }
+    expect(error).toEqual(undefined);
+  });
+
+  it("renders correct snapshot", () => {
+    const wrapper = mount(<Toolbar />);
+
+    expect(toJson(wrapper)).toMatchSnapshot();
+  });
```
jordanell, line 23: Just a shout out to @licarm , Imo and you have been doing the snapshot test slightly differently from PR to PR. Sometimes we put this as part of the default props tests, and sometimes we make it a stand alone test like this.

We should be consistent. My 2 cents is that I like it as a stand alone test like this.

packages/riipen-ui/src/components/Toolbar.test.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+  describe("children prop", () => {
+    it("renders given children inside the div element", () => {
+      const child = <h5>hello</h5>;
```
jordanell, line 27: Is it me you're looking for?

jordanell (approved):
Great job!

packages/riipen-ui/src/components/Toolbar.test.jsx
imogenem: We'll switch to a stand alone snapshot test!

PR #126, opened 2021-02-02
[PLT-6648] Unit tests for List

imogenem:
## Description
Unit tests for List component

## Where to Start
List.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/List.test.jsx
+++ b/packages/riipen-ui/src/components/List.test.jsx
@@ -1,0 +1,61 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import List from "./List";
+
+describe("<List>", () => {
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/List.test.jsx
```diff
@@ -0,0 +1,61 @@
...
+    expect(error).toEqual(undefined);
+  });
+
+  describe("snapshot test", () => {
+    it("renders correctly", () => {
+      const wrapper = mount(<List />);
+
+      expect(toJson(wrapper)).toMatchSnapshot();
+    });
+  });
```
jordanell, line 26: @imogenem @licarm another example of the snapshot test. Let's update this to be whatever we choose.

jordanell (changes requested):
Will approve once we figure out how we want to do the snapshot tests.

jordanell (approved):

PR #133, opened 2021-02-03
[PLT-6643] tests for Hidden component

licarm:
## Description
- tests for Hidden component
- clarified description for the direction prop

## Notes
Interested to see if classes prop tests are okay because there isn't an explicit classes prop, even though it still allows classes to be passed in.

### DIFF
--- a/packages/riipen-ui/src/components/Hidden.test.jsx
+++ b/packages/riipen-ui/src/components/Hidden.test.jsx
@@ -58,7 +58,6 @@ describe("<Hidden>", () => {
     it("applies given direction to the size-direction className on div", () => {
       const direction = "up";
       const size = "sm";
-      const className = new RegExp(`.+-${direction}`);
 
       const wrapper = mount(<Hidden direction={direction} size={size} />);
 
@@ -66,7 +65,7 @@ describe("<Hidden>", () => {
         wrapper
           .find("Hidden")
           .find("div")
-          .hasClass(className)
+          .hasClass(new RegExp(`.+-${direction}`))
       ).toEqual(true);
     });
 
@@ -85,7 +84,6 @@ describe("<Hidden>", () => {
   describe("size prop", () => {
     it("applies given size to the size-direction className on div", () => {
       const size = "sm";
-      const className = new RegExp(`${size}-.+`);
 
       const wrapper = mount(<Hidden size={size} />);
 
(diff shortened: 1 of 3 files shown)


packages/riipen-ui/src/components/Hidden.test.jsx
```diff
@@ -0,0 +1,125 @@
...
+          .contains(child)
+      ).toEqual(true);
+    });
+  });
+
+  describe("direction prop", () => {
+    it("applies given direction to the size-direction className on div", () => {
+      const direction = "up";
+      const size = "sm";
+      const className = new RegExp(`.+-${direction}`);
```
jordanell, line 61: I would maybe argue we don't need this in a variable and that we could just inline this value on line 69 below. The reason being we basically inline every other test outcome so might as well continue that trend.

packages/riipen-ui/src/components/Hidden.test.jsx
```diff
@@ -0,0 +1,125 @@
...
+      mount(<Hidden size={size} direction={direction} />);
+
+      expect(errors).toHaveBeenCalled();
+    });
+  });
+
+  describe("size prop", () => {
+    it("applies given size to the size-direction className on div", () => {
+      const size = "sm";
+      const className = new RegExp(`${size}-.+`);
```
jordanell, line 88: Same thing here with inlining.

packages/riipen-ui/src/components/Hidden.test.jsx
```diff
@@ -0,0 +1,125 @@
...
+      const wrapper = mount(<Hidden size={size} />);
+
+      expect(
+        wrapper
+          .find("Hidden")
+          .props()
+          .classes.sort()
+      ).toEqual(classes.sort());
+    });
+  });
```
jordanell, line 124: I think this actually uncovered a source issue. We aren't actually using the `classes` prop anywhere.  We should either:

1. Add the `classes` prop to this component just like all others
2. Remove this test because it doesn't actually test anything in the actual component itself

I think we should do 2 because this component really would have no reason to have additional classes applied to it. But I am open to suggestions.

jordanell (changes requested):

packages/riipen-ui/src/components/Hidden.test.jsx
licarm: I'm going with option 2!

jordanell (approved):

PR #134, opened 2021-02-03
[PLT-6660] Unit Tests for Spinner

imogenem:
## Description
Unit Tests for Spinner

## Where to Start
Spinner.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/Spinner.test.jsx
+++ b/packages/riipen-ui/src/components/Spinner.test.jsx
@@ -1,0 +1,98 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import Spinner from "./Spinner";
+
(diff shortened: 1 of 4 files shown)


packages/riipen-ui/src/components/Spinner.test.jsx
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  describe("children prop", () => {
+    it("displays given children", () => {
+      const child = <div>child</div>;
+
+      const wrapper = mount(<Spinner>{child}</Spinner>);
+
+      expect(wrapper.containsMatchingElement(child)).toBeTruthy();
+    });
```
jordanell, line 41: Maybe add one more test here: "does not display children when loading is true".

jordanell (changes requested):

jordanell (approved):

PR #135, opened 2021-02-04
[PLT-6656] Unit tests for Radio

imogenem:
## Description
Unit tests for Radio

## Notes
Tests are passing, but need to look in to the warning message that appears to relate to RadioGroup's onChange prop. The warning message notes that the checked prop appears to require an onChange handler when used.

## Where to Start
Radio.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/Radio.test.jsx
+++ b/packages/riipen-ui/src/components/Radio.test.jsx
@@ -9,7 +9,7 @@ describe("<Radio>", () => {
     let error;
 
     try {
-      mount(<Radio />);
+      mount(<Radio readOnly />);
     } catch (theError) {
       error = theError;
     }
@@ -18,7 +18,7 @@ describe("<Radio>", () => {
   });
 
   it("renders correct snapshot", () => {
-    const wrapper = mount(<Radio />);
+    const wrapper = mount(<Radio readOnly />);
 
     expect(toJson(wrapper)).toMatchSnapshot();
   });
@@ -27,7 +27,7 @@ describe("<Radio>", () => {
     it("sets correct default props", () => {
       const defaultProps = new Radio().type.defaultProps;
 
-      const wrapper = mount(<Radio />);
+      const wrapper = mount(<Radio readOnly />);
 
       expect(wrapper.find("Radio").props().checked).toEqual(
         defaultProps.checked
@@ -42,7 +42,7 @@ describe("<Radio>", () => {
     it("sets valid custom checked", () => {
       const checked = true;
 
-      const wrapper = mount(<Radio checked={checked} />);
+      const wrapper = mount(<Radio checked={checked} readOnly />);
 
       expect(
         wrapper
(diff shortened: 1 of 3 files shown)


packages/riipen-ui/src/components/Radio.test.jsx
```diff
@@ -0,0 +1,152 @@
...
+          .hasClass(color)
+      ).toBeTruthy();
+      expect(
+        wrapper
+          .find("Radio")
+          .find("span")
+          .hasClass("disabled")
+      ).toBeFalsy();
+    });
+  });
```
licarm, line 136: I think you meant to check `const disabled = false;`, not color.

licarm (changes requested):

packages/riipen-ui/src/components/Radio.test.jsx
```diff
@@ -0,0 +1,152 @@
...
+
+      const wrapper = mount(<Radio checked={checked} />);
+
+      expect(
+        wrapper
+          .find("Radio")
+          .find("input")
+          .props().checked
+      ).toEqual(checked);
+    });
```
jordanell, line 53: I am guessing this is throwing the warning:

```
Warning: Failed prop type: You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`.
```

To make this go away you could do `<Radio checked={checked} readOnly />`.

What's happening is that if you have an HTML checkbox and you say "this checkbox is checked", you basically can never uncheck it, so it's telling you to pass "readOnly" to the checkbox so that the browser knows that this checkbox will never be unchecked.

packages/riipen-ui/src/components/Radio.test.jsx
```diff
@@ -0,0 +1,152 @@
...
+
+      expect(errors).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe("disabled prop", () => {
+    it("renders radio as disabled when disabled prop is true", () => {
+      const disabled = true;
+
+      const wrapper = mount(<Radio disabled={disabled} />);
```
jordanell, line 108: You might have the readonly warning here again, but I am not 100% sure.

jordanell (approved):
Going to approve as I'm not sure the comments I made are valid, but we should fix the things I pointed out if they are not valid.

packages/riipen-ui/src/components/Radio.test.jsx
imogenem: disabled is set to false by default, so with this test I was trying to target line 90 in the component file.

```
 className={clsx("checkmark", disabled ? "disabled" : null, color)}
```

packages/riipen-ui/src/components/Radio.test.jsx
imogenem: The warning was affecting every test, so I've adjusted every mount function

PR #136, opened 2021-02-04
[PLT-6670] tests for typography component

licarm:
## Where to Start
Typography.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/Typography.test.jsx
+++ b/packages/riipen-ui/src/components/Typography.test.jsx
@@ -3,7 +3,6 @@ import toJson from "enzyme-to-json";
 import React from "react";
 
 import Typography from "./Typography";
-// import { variantMapping } from "./Typography";
 
 describe("<Typography>", () => {
   it("renders without errors", () => {
--- a/packages/riipen-ui/src/components/Typography.test.jsx
+++ b/packages/riipen-ui/src/components/Typography.test.jsx
@@ -1,0 +1,289 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import Typography from "./Typography";
+// import { variantMapping } from "./Typography";
+
+describe("<Typography>", () => {
+  it("renders without errors", () => {
+    let error;
+
+    try {
+      mount(<Typography />);
+    } catch (theError) {
+      error = theError;
+    }
+
+    expect(error).toEqual(undefined);
+  });
+
+  it("renders correct snapshot", () => {
+    const wrapper = mount(<Typography />);
+
+    expect(toJson(wrapper)).toMatchSnapshot();
+  });
+
+  describe("default props", () => {
+    it("sets correct default props", () => {
+      const defaultProps = new Typography().type.defaultProps;
+
+      const wrapper = mount(<Typography />);
+
+      const component = wrapper.find("Typography");
+      expect(component.props().color).toEqual(defaultProps.color);
+      expect(component.props().display).toEqual(defaultProps.display);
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/Typography.test.jsx
```diff
@@ -0,0 +1,289 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import Typography from "./Typography";
+// import { variantMapping } from "./Typography";
```
imogenem, line 6: Is this needed if it's commented out?

packages/riipen-ui/src/components/Typography.test.jsx
```diff
@@ -0,0 +1,289 @@
...
+
+      expect(
+        wrapper
+          .find("Typography")
+          .props()
+          .classes.sort()
+      ).toEqual(classes.sort());
+    });
+
+    it("gives an error when classes are provided as one string", () => {
```
imogenem, line 79: Is this test necessary since we stopped using the test where two classes are added in the array?

imogenem (commented):
Superb! Tres bien!

packages/riipen-ui/src/components/Typography.test.jsx
licarm: Ahh it's because html classes are commonly added as a string, like 
`<div class="one two three"></div>`.

packages/riipen-ui/src/components/Typography.test.jsx
licarm: thanks!

packages/riipen-ui/src/components/Typography.test.jsx
jordanell: I mean, we _probably_ don't need this test. This test is actually testing to make sure `PropTypes` are working.

1. We have a prop of `classes` that has to be an array
2. We are trying to pass a string instead of array
3. We expect and error

This is more so testing React's internals rather than any code we have written.

packages/riipen-ui/src/components/Typography.test.jsx
```diff
@@ -0,0 +1,288 @@
...
+    it("sets component with valid custom component", () => {
+      const component = "a";
+
+      const wrapper = mount(<Typography component={component} />);
+
+      expect(
+        wrapper
+          .find("Typography")
+          .childAt(0)
+          .name()
```
jordanell, line 122: Nice!

packages/riipen-ui/src/components/Typography.test.jsx
```diff
@@ -0,0 +1,288 @@
...
+    it("gives an error when given an invalid component", () => {
+      const component = "invalidcomponent";
+      const errors = jest.spyOn(console, "error").mockImplementation();
+
+      mount(<Typography component={component} />);
+
+      expect(errors).toHaveBeenCalled();
+    });
+
+    it("applies the component prop over the variant prop when both are provided", () => {
```
jordanell, line 135: Perfect!

packages/riipen-ui/src/components/Typography.test.jsx
```diff
@@ -0,0 +1,288 @@
...
+
+      expect(
+        wrapper
+          .find("Typography")
+          .childAt(0)
+          .name()
+      ).toEqual(variant);
+    });
+
+    it("sets component to span when given an unmapped variant", () => {
```
jordanell, line 266: Nice test!

jordanell (approved):

PR #137, opened 2021-02-04
[PLT-6668] tests for Tabs component

licarm:
## Where to Start
Tabs.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/Tabs.test.jsx
+++ b/packages/riipen-ui/src/components/Tabs.test.jsx
@@ -1,0 +1,233 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import Tab from "./Tab";
+import Tabs from "./Tabs";
+
+describe("<Tabs>", () => {
+  it("renders without errors", () => {
+    let error;
+
+    try {
+      mount(<Tabs />);
+    } catch (theError) {
+      error = theError;
+    }
+
+    expect(error).toEqual(undefined);
+  });
+
+  it("renders correct snapshot", () => {
+    const wrapper = mount(<Tabs />);
+
+    expect(toJson(wrapper)).toMatchSnapshot();
+  });
+
+  describe("default props", () => {
+    it("sets correct default props", () => {
+      const defaultProps = new Tabs().type.defaultProps;
+
+      const wrapper = mount(<Tabs />);
+
+      const component = wrapper.find("Tabs");
+      expect(component.props().breakpoint).toEqual(defaultProps.breakpoint);
+      expect(component.props().color).toEqual(defaultProps.color);
+      expect(component.props().component).toEqual(defaultProps.component);
+      expect(component.props().orientation).toEqual(defaultProps.orientation);
+      expect(component.props().variant).toEqual(defaultProps.variant);
+    });
+  });
+
+  describe("breakpoint prop", () => {
+    it("passes valid given breakpoint to children nodes", () => {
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/Tabs.test.jsx
```diff
@@ -0,0 +1,233 @@
...
+    });
+  });
+
+  describe("component prop", () => {
+    it("applies given component to the root node", () => {
+      const component = "h3";
+
+      const wrapper = mount(<Tabs component={component} />);
+
+      expect(wrapper.find("Tabs").childAt());
```
imogenem, line 142: Is this supposed to be....

```
     expect(
        wrapper
          .find("Tabs")
          .childAt(0)
          .name()
      ).toEqual(component);
```

packages/riipen-ui/src/components/Tabs.test.jsx
```diff
@@ -0,0 +1,233 @@
...
+      const wrapper = mount(
+        <Tabs value={value}>
+          {child1}
+          {child2}
+        </Tabs>
+      );
+
+      const childrenNodes = wrapper.find("Tab");
+      expect(childrenNodes.at(0).props().active).toEqual(false);
+      expect(childrenNodes.at(1).props().active).toEqual(true);
```
imogenem, line 202: Coooool

packages/riipen-ui/src/components/Tabs.test.jsx
```diff
@@ -0,0 +1,233 @@
...
+      const wrapper = mount(
+        <Tabs>
+          {child1}
+          {child2}
+        </Tabs>
+      );
+
+      const childrenNodes = wrapper.find("Tab");
+      expect(childrenNodes.at(0).props().value).toEqual("one");
+      expect(childrenNodes.at(1).props().value).toEqual("two");
```
imogenem, line 76: Is the second test necessary here? Looks like they check the same thing.

imogenem (commented):
Nice use of Tab!

packages/riipen-ui/src/components/Tabs.test.jsx
```diff
@@ -0,0 +1,233 @@
...
+      const component = "abcd";
+
+      mount(<Tabs component={component} />);
+
+      expect(errors).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe("onChange prop", () => {
+    it("passes valid handler to the onClick prop of children nodes", () => {
```
jordanell, line 156: You are mixing your tests up here I think. You are saying "check to make sure it's passed to children", but then your test is actually testing the click functionality.

I think you should maybe split this into two tests:
1. Does it pass the prop to the children
2. Does the click functionality work

packages/riipen-ui/src/components/Tabs.test.jsx
```diff
@@ -0,0 +1,233 @@
...
+      const wrapper = mount(<Tabs onChange={onChange}>{child}</Tabs>);
+
+      wrapper.find("Tab").simulate("click");
+
+      expect(onChange).toHaveBeenCalledWith(expect.any(Object), "one");
+    });
+  });
+
+  describe("orientation prop", () => {
+    it("passes given valid orientation to children", () => {
```
jordanell, line 169: I think we should add a second test that checks for the `orientation` prop is also used in the `className` found on line 94.

jordanell (changes requested):

jordanell (approved):

PR #138, opened 2021-02-05
[PLT-6659] tests for RadioGroup component

licarm:
## Description
 tests for RadioGroup component

### DIFF
--- a/packages/riipen-ui/src/components/RadioGroup.jsx
+++ b/packages/riipen-ui/src/components/RadioGroup.jsx
@@ -69,9 +69,11 @@ class RadioGroup extends React.Component {
       ...other
     } = this.props;
 
-    const childrenWithProps = React.Children.map(children, child =>
-      React.cloneElement(child, { onChange: this.handleChange })
-    );
+    const childrenWithProps = React.Children.map(children, child => {
+      if (!child) return null;
+
+      return React.cloneElement(child, { onChange: this.handleChange });
+    });
 
     return (
       <React.Fragment>
--- a/packages/riipen-ui/src/components/RadioGroup.test.jsx
+++ b/packages/riipen-ui/src/components/RadioGroup.test.jsx
@@ -1,0 +1,153 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import RadioGroup from "./RadioGroup.jsx";
+import Radio from "./Radio.jsx";
+
+describe("<RadioGroup>", () => {
+  it("renders without errors", () => {
+    let error;
+
+    try {
+      mount(<RadioGroup />);
+    } catch (theError) {
+      error = theError;
+    }
+
+    expect(error).toEqual(undefined);
+  });
+
+  it("renders correct snapshot", () => {
+    const wrapper = mount(<RadioGroup />);
+
+    expect(toJson(wrapper)).toMatchSnapshot();
+  });
+
+  describe("children prop", () => {
+    it("renders given non null children", () => {
+      const child = <Radio label="one" readOnly />;
+      const child2 = <Radio label="two" readOnly />;
+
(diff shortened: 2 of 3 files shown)


packages/riipen-ui/src/components/RadioGroup.test.jsx
```diff
@@ -0,0 +1,153 @@
...
+
+      const wrapper = mount(
+        <RadioGroup onChange={onChange}>{child}</RadioGroup>
+      );
+
+      wrapper
+        .find("Radio")
+        .find("input")
+        .simulate("change");
+
```
imogenem, line 119: Should there also be a test that covers the cloneElement function in lines 72-73 in RadioGroup? 
ie. something like...
expect(wrapper.find("fieldset").contains(childrenWithProps)).toEqual(true);

packages/riipen-ui/src/components/RadioGroup.test.jsx
```diff
@@ -0,0 +1,153 @@
...
+      expect(wrapper.find("InputLabel")).toHaveLength(1);
+    });
+
+    it("passes given label to InputLabel as children", () => {
+      const label = "it's a label";
+
+      const wrapper = mount(<RadioGroup label={label} />);
+
+      expect(wrapper.find("InputLabel").props().children).toEqual(label);
+    });
```
imogenem, line 89: Since these tests use the same setup, is it worthwhile to combine them?

Something like...
"given label is rendered as InputLabel and passed to InputLabel as children"

packages/riipen-ui/src/components/RadioGroup.test.jsx
```diff
@@ -0,0 +1,153 @@
...
+
+    it("does not render InputLabel when both label and hint are not provided", () => {
```
imogenem, line 91: Should this just be for when label is not provided? It looks like it can render without a hint based on the tests just above this.

imogenem (changes requested):
Looks great!

packages/riipen-ui/src/components/RadioGroup.jsx
```diff
@@ -69,9 +69,11 @@ class RadioGroup extends React.Component {
...
     } = this.props;
 
-    const childrenWithProps = React.Children.map(children, child =>
-      React.cloneElement(child, { onChange: this.handleChange })
-    );
+    const childrenWithProps = React.Children.map(children, child => {
+      if (!child) return null;
+
+      return React.cloneElement(child, { onChange: this.handleChange });
+    });
```
jordanell, line 76: Perfect!

packages/riipen-ui/src/components/RadioGroup.test.jsx
jordanell: I think we are already testing that with the `children` tests.

packages/riipen-ui/src/components/RadioGroup.test.jsx
```diff
@@ -0,0 +1,153 @@
...
+    });
+
+    it("sets required prop of InputLabel to false when required is false", () => {
+      const label = "it's a label";
+      const required = false;
+
+      const wrapper = mount(<RadioGroup label={label} required={required} />);
+
+      expect(wrapper.find("InputLabel").props().required).toEqual(required);
+    });
```
jordanell, line 141: I don't think you need both of these. You could just have a test "passes required prop to InputLabel" or something.

jordanell (changes requested):

packages/riipen-ui/src/components/RadioGroup.test.jsx
licarm: InputLabel will be displayed if either hint or label (or both) are provided

jordanell (approved):

imogenem (approved):

PR #139, opened 2021-02-08
[PLT-6661] Unit Tests for Tab

imogenem:
## Description
Unit Tests for Tab

## Notes
Default props are not formatted the same as the other components, had to adjust that test.

## Where to Start
Tab.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/Tab.jsx
+++ b/packages/riipen-ui/src/components/Tab.jsx
@@ -228,7 +228,7 @@ Tab.propTypes = {
   icon: PropTypes.elementType,
 
   /**
-   * The label element.
+   * The label element
    */
   label: PropTypes.node,
 
--- a/packages/riipen-ui/src/components/Tab.jsx
+++ b/packages/riipen-ui/src/components/Tab.jsx
@@ -213,12 +213,12 @@ Tab.propTypes = {
   disabled: PropTypes.bool,
 
   /**
-   * If `true` will force the tab to be displayed active
+   * If `true` will force the tab to be displayed active.
    */
   displayActive: PropTypes.bool,
 
   /**
-   * If `true`, will make the tab grow to use all the available space,
+   * If `true`, will make the tab grow to use all the available space.
    */
   fullWidth: PropTypes.bool,
 
@@ -228,15 +228,15 @@ Tab.propTypes = {
   icon: PropTypes.elementType,
 
   /**
-   * The label element
+   * The label element.
    */
   label: PropTypes.node,
 
   /**
    * Callback fired when the tab is clicked.
    *
-   * @param {object} event The event source of the callback
-   * @param {any} value We default to the index of the child (number)
+   * @param {object} event The event source of the callback.
+   * @param {any} value We default to the index of the child (number).
    */
   onClick: PropTypes.func,
 
@@ -247,7 +247,7 @@ Tab.propTypes = {
 
   /**
    * @ignore
-   * The theme context object
+   * The theme context object.
    */
   theme: PropTypes.object,
 
--- a/packages/riipen-ui/src/components/Tab.test.jsx
+++ b/packages/riipen-ui/src/components/Tab.test.jsx
@@ -2,6 +2,11 @@ import { mount } from "enzyme";
 import toJson from "enzyme-to-json";
 import React from "react";
 
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faPhone } from "@fortawesome/free-solid-svg-icons";
+
+const icon = i => props => <FontAwesomeIcon icon={i} {...props} />;
+
 import Tab from "./Tab";
 
 describe("<Tab>", () => {
@@ -9,7 +14,7 @@ describe("<Tab>", () => {
     let error;
 
     try {
-      mount(<Tab />);
+      mount(<Tab value />);
     } catch (theError) {
       error = theError;
     }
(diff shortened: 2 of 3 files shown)


licarm (approved):
Nice!! :)

packages/riipen-ui/src/components/Tab.test.jsx
```diff
@@ -0,0 +1,363 @@
...
+      expect(wrapper.find("Tab").props().active).toEqual(false);
+      expect(wrapper.find("Tab").props().breakpoint).toEqual("sm");
+      expect(wrapper.find("Tab").props().color).toEqual("secondary");
+      expect(wrapper.find("Tab").props().disabled).toEqual(false);
+      expect(wrapper.find("Tab").props().fullWidth).toEqual(false);
+      expect(wrapper.find("Tab").props().orientation).toEqual("horizontal");
+    });
+  });
+
+  describe("active and displayActive props", () => {
```
jordanell, line 44: There's a lot going on in this set of tests which I think we could simplify or spread out a bit. A general rule of writing tests is "don't make the next person think" when they have to read / update your tests, and I had to do more thinking than usual with these ones.

There couple things that are going on here:

1. We want to make sure that the correct class name, based on color, is set when active is true
2. We want to make sure that the correct class name, based on color, is set when displayActive is true
3. We want to make sure that the correct class name, based on orientation, is set when active is true
4. We want to make sure that the correct class name, based on orientation, is set when displayActive is true

Those being our cases, I think we can split this one giant describe block out into each prop's unique describe block.

- Case 1 above could be part of the `active` prop tests.
- Case 2 above could be part of the `displayActive` prop tests.
- Case 3 above could be part of the `active` prop tests.
- Case 4 above could be part of the `displayActive` prop tests.

We can also simplify the tests in general:

- We probably don't need to test every single color / orientation, just one is probably good enough for each case I just described.
- We don't need to test `active` and `displayActive` in tandum with each other since they actually don't affect each other

packages/riipen-ui/src/components/Tab.test.jsx
```diff
@@ -0,0 +1,363 @@
...
+  describe("breakpoint prop", () => {
+    describe("custom breakpoint and orientation props", () => {
+      const props = {};
+
+      const tests = [
+        {
+          breakpoint: "md",
+          orientation: "horizontal",
+          argument: "breakpoint",
+          expected: true
```
jordanell, line 128: I think you could maybe make this a bit simpler in tandem with what is below on lines 146-147. What if you defined it like:

```js
{
  props: {
    breakpoint: 'md',
    orientation: 'horizontal',
  },
  expected: true,
}
```

then  your test below could be:

```js
tests.forEach(test => {
  it(`${test.expected ? 'does' : 'does not'} set breakpoint class name with breakpoint of ${test.props.breakpoint} and orientation of ${test.props.orientation}`, () => {
    const wrapper = mount(<Tab {...test.props} value />);

    expect(
      wrapper
        .find("Tab")
        .childAt(0)
        .hasClass("breakpoint")
    ).toEqual(test.expected);
  });
})
```

packages/riipen-ui/src/components/Tab.test.jsx
```diff
@@ -0,0 +1,363 @@
...
+        wrapper
+          .find("Tab")
+          .childAt(0)
+          .childAt(0)
+          .hasClass("icon")
+      ).toEqual(true);
+    });
+  });
+
+  describe("label prop", () => {
```
jordanell, line 254: I would maybe add a test to make sure it renders the label value as well.

packages/riipen-ui/src/components/Tab.test.jsx
```diff
@@ -0,0 +1,363 @@
...
+          .find("Tab")
+          .childAt(0)
+          .childAt(0)
+          .hasClass("label")
+      ).toEqual(true);
+    });
+  });
+
+  describe("onClick prop", () => {
+    it("onClick prop correctly handles click event", () => {
```
jordanell, line 271: I would be slightly more descriptive with this test name. What if you changed it to "invokes onClick prop on click" That way you know that "correct" means it is called. See below why that might be good.

packages/riipen-ui/src/components/Tab.test.jsx
```diff
@@ -0,0 +1,363 @@
...
+      const handler = jest.fn();
+
+      const wrapper = mount(<Tab onClick={handler} value />);
+      wrapper.simulate("click");
+
+      expect(handler).toHaveBeenCalledTimes(1);
+    });
+
+    it("onClick prop correctly handles click event when disabled", () => {
```
jordanell, line 280: Then here, you can say "does not invoke onClick prop on click when disabled". Now it's clear what the behaviour should be without reading the test itself.

packages/riipen-ui/src/components/Tab.test.jsx
```diff
@@ -0,0 +1,363 @@
...
+
+      const wrapper = mount(
+        <Tab onClick={handler} disabled={disabled} value />
+      );
+      wrapper.simulate("click");
+
+      expect(handler).toHaveBeenCalledTimes(0);
+    });
+
+    it("onClick prop correctly handles valid keydown event", () => {
```
jordanell, line 292: Same here for naming.

packages/riipen-ui/src/components/Tab.test.jsx
```diff
@@ -0,0 +1,363 @@
...
+      const handler = jest.fn();
+
+      const wrapper = mount(<Tab onClick={handler} value />);
+      wrapper.simulate("keydown", { key: "Enter" });
```
jordanell, line 296: Nice!

packages/riipen-ui/src/components/Tab.test.jsx
```diff
@@ -0,0 +1,363 @@
...
+
+      expect(handler).toHaveBeenCalledTimes(1);
+    });
+
+    it("onClick prop correctly handles invalid keydown event", () => {
```
jordanell, line 301: Same here for naming.

packages/riipen-ui/src/components/Tab.test.jsx
```diff
@@ -0,0 +1,363 @@
...
+      const errors = jest.spyOn(console, "error").mockImplementation();
+      const orientation = "upright";
+
+      mount(<Tab orientation={orientation} value />);
+
+      expect(errors).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe("focusVisible state", () => {
```
jordanell, line 335: Nice!

jordanell (changes requested):

packages/riipen-ui/src/components/Tab.jsx
```diff
@@ -228,7 +228,7 @@ Tab.propTypes = {
   icon: PropTypes.elementType,
 
   /**
-   * The label element.
```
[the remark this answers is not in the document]
imogenem: adjusted prop descriptions

packages/riipen-ui/src/components/Tab.test.jsx
imogenem: Added

packages/riipen-ui/src/components/Tab.test.jsx
imogenem: Makes sense

packages/riipen-ui/src/components/Tab.test.jsx
imogenem: Adjusted, I like the use of the ternary operator in the test name

jordanell (approved):
Nice job!

PR #142, opened 2021-02-09
[PLT-6665] tests for TableHeader component

licarm:
## Where to Start
TableHeader.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/TableHeader.test.jsx
+++ b/packages/riipen-ui/src/components/TableHeader.test.jsx
@@ -1,0 +1,88 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import Table from "./Table.jsx";
+import TableHeader from "./TableHeader.jsx";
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/TableHeader.test.jsx
```diff
@@ -0,0 +1,88 @@
...
+        wrapper
+          .find("thead")
+          .childAt(0)
+          .matchesElement(child)
+      ).toEqual(true);
+    });
+  });
+
+  describe("classes prop", () => {
+    it("applies classes to the root node of Typography", () => {
```
imogenem, line 59: Should this just be "applies classes to the root node" since Typography isn't referenced in the tests?

jordanell (changes requested):
Same comment as @imogenem

packages/riipen-ui/src/components/TableHeader.test.jsx
licarm: oooops

imogenem (approved):

PR #145, opened 2021-02-09
[PLT-6657] Unit Tests for RadioButton

imogenem:
## Description
Unit Tests for RadioButton

## Where to Start
RadioButton.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/RadioButton.test.jsx
+++ b/packages/riipen-ui/src/components/RadioButton.test.jsx
@@ -1,0 +1,154 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import { faFlag } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+import RadioButton from "./RadioButton";
+
+const icon = i => <FontAwesomeIcon icon={i} />;
+
+describe("<RadioButton>", () => {
+  it("renders without errors", () => {
+    let error;
+
+    try {
+      mount(<RadioButton />);
+    } catch (theError) {
+      error = theError;
+    }
+
+    expect(error).toEqual(undefined);
+  });
+
+  it("renders correct snapshot", () => {
+    const wrapper = mount(<RadioButton />);
+
+    expect(toJson(wrapper)).toMatchSnapshot();
+  });
+
+  describe("checked prop", () => {
+    it("sets class name as checked and color as initial when checked is true", () => {
+      const checked = true;
+
+      const wrapper = mount(<RadioButton checked={checked} readOnly />);
+
+      expect(
+        wrapper
+          .find("RadioButton")
+          .childAt(0)
+          .hasClass("checked")
+      ).toEqual(true);
+      expect(wrapper.find("Typography").props().color).toEqual("initial");
+    });
+
+    it("sets color as grey600 when checked is false", () => {
+      const checked = false;
+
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/RadioButton.test.jsx
```diff
@@ -0,0 +1,154 @@
...
+      const wrapper = mount(<RadioButton checked={checked} readOnly />);
+
+      expect(
+        wrapper
+          .find("RadioButton")
+          .childAt(0)
+          .hasClass("checked")
+      ).toEqual(true);
+      expect(wrapper.find("Typography").props().color).toEqual("initial");
+    });
```
licarm, line 44: checked is also used in the label and input elements. Maybe we should test those too?

packages/riipen-ui/src/components/RadioButton.test.jsx
```diff
@@ -0,0 +1,154 @@
...
+      expect(
+        wrapper
+          .find("RadioButton")
+          .props()
+          .classes.sort()
+      ).toEqual(classes.sort());
+    });
+  });
+
+  describe("disabled prop", () => {
```
licarm, line 83: Same question as checked prop. Disabled prop is also used in the label and input elements.

packages/riipen-ui/src/components/RadioButton.test.jsx
jordanell: It's also used in `<input>` so we should probably add a test for that as well.

jordanell (changes requested):

packages/riipen-ui/src/components/RadioButton.test.jsx
imogenem: Adjusted, let me know what you think of the test name.

packages/riipen-ui/src/components/RadioButton.test.jsx
```diff
@@ -0,0 +1,158 @@
...
+      expect(
+        wrapper
+          .find("RadioButton")
+          .childAt(0)
+          .hasClass("checked")
+      ).toEqual(true);
+      expect(wrapper.find("label").hasClass("checked")).toEqual(true);
+      expect(wrapper.find("input").props().checked).toEqual(checked);
+      expect(wrapper.find("Typography").props().color).toEqual("initial");
+    });
```
jordanell, line 46: I am actually wondering if this should be 4 separate tests. The test name no longer reflects everything that is being tested here. 

As a rule of thumb, it's generally a good idea to only actually test 1 thing per test. Reason being, if this test fails the test name and what causes it to fail now no longer align and it can be hard to debug. Also, the first expect might fail causing the others to actually not run. So you go and fix the first expect, but then the 2nd fails, and so on and so on. So you actually had only 1 test failure, but more than 1 thing was failing.

packages/riipen-ui/src/components/RadioButton.test.jsx
```diff
@@ -0,0 +1,158 @@
...
+
+      expect(
+        wrapper
+          .find("RadioButton")
+          .childAt(0)
+          .hasClass("disabled")
+      ).toEqual(true);
+      expect(wrapper.find("label").hasClass("disabled")).toEqual(true);
+      expect(wrapper.find("input").props().disabled).toEqual(disabled);
+    });
```
jordanell, line 99: Same here for multiple tests.

jordanell (changes requested):

jordanell (approved):

PR #146, opened 2021-02-10
[Plt-6621] tests for Accordion component

licarm:
## Description

- tests for the Accordion component
- focus, blur, focusVisible removed from AccordionSummary

### DIFF
--- a/packages/riipen-ui/src/components/Accordion.test.jsx
+++ b/packages/riipen-ui/src/components/Accordion.test.jsx
@@ -9,10 +9,13 @@ import AccordionSummary from "./AccordionSummary";
 describe("<Accordion>", () => {
   it("renders without errors", () => {
     let error;
-    const child = <AccordionDetails />;
 
     try {
-      mount(<Accordion>{child}</Accordion>);
+      mount(
+        <Accordion>
+          <AccordionDetails />
+        </Accordion>
+      );
     } catch (e) {
       error = e;
     }
@@ -21,13 +24,10 @@ describe("<Accordion>", () => {
   });
 
   it("renders correct snapshot", () => {
-    const summary = <AccordionSummary />;
-    const child = <AccordionDetails />;
-
     const wrapper = mount(
       <Accordion>
-        {summary}
-        {child}
+        <AccordionSummary />
+        <AccordionDetails />
       </Accordion>
     );
 
@@ -36,9 +36,11 @@ describe("<Accordion>", () => {
 
   describe("default props", () => {
     it("sets correct default props", () => {
-      const child = <AccordionDetails />;
-
-      const wrapper = mount(<Accordion>{child}</Accordion>);
+      const wrapper = mount(
+        <Accordion>
+          <AccordionDetails />
+        </Accordion>
+      );
 
       expect(wrapper.find("Accordion").props().classes).toEqual([]);
       expect(wrapper.find("Accordion").props().defaultExpanded).toEqual(false);
@@ -52,18 +54,19 @@ describe("<Accordion>", () => {
 
       const wrapper = mount(<Accordion>{child}</Accordion>);
 
-      expect(wrapper.find("AccordionDetails").props().children).toEqual(
-        "hello"
-      );
+      expect(wrapper.find("Accordion").contains(child)).toEqual(true);
     });
   });
 
   describe("classes prop", () => {
     it("applies custom classes", () => {
       const classes = ["classOne"];
-      const child = <AccordionDetails>hello</AccordionDetails>;
 
-      const wrapper = mount(<Accordion classes={classes}>{child}</Accordion>);
+      const wrapper = mount(
+        <Accordion classes={classes}>
+          <AccordionDetails>hello</AccordionDetails>
+        </Accordion>
+      );
 
       expect(
         wrapper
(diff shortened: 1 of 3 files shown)


packages/riipen-ui/src/components/Accordion.test.jsx
```diff
@@ -0,0 +1,149 @@
...
+        wrapper
+          .find("Accordion")
+          .childAt(0)
+          .hasClass("expanded")
+      ).toEqual(false);
+    });
+  });
+
+  describe("disabled prop", () => {
+    it("applies disabled class when true", () => {
```
imogenem, line 121: Just to make the test name really clear, should this be "applies disabled class name when disabled is true"

packages/riipen-ui/src/components/Accordion.test.jsx
```diff
@@ -0,0 +1,149 @@
...
+
+      expect(
+        wrapper
+          .find("Accordion")
+          .childAt(0)
+          .hasClass("disabled")
+      ).toEqual(true);
+    });
+
+    it("does not apply disabled class when false", () => {
```
imogenem, line 135: Same change needed to test name

packages/riipen-ui/src/components/Accordion.test.jsx
```diff
@@ -0,0 +1,149 @@
...
+
+    expect(toJson(wrapper)).toMatchSnapshot();
+  });
+
+  describe("default props", () => {
+    it("sets correct default props", () => {
+      const child = <AccordionDetails />;
+
+      const wrapper = mount(<Accordion>{child}</Accordion>);
+
```
imogenem, line 42: I was able to test the default classes value for Button, does this work for you?

```
 expect(wrapper.find("Accordion").props().classes).toEqual([]);
```

imogenem (changes requested):
Just small changes, looks great

packages/riipen-ui/src/components/Accordion.test.jsx
```diff
@@ -0,0 +1,150 @@
...
+    } catch (e) {
+      error = e;
+    }
+
+    expect(error).toEqual(undefined);
+  });
+
+  it("renders correct snapshot", () => {
+    const summary = <AccordionSummary />;
+    const child = <AccordionDetails />;
```
jordanell, line 25: I don't think these need to be variables. I would just render them inline on lines 29-30.

packages/riipen-ui/src/components/Accordion.test.jsx
```diff
@@ -0,0 +1,150 @@
...
+    it("renders given children", () => {
+      const child = <AccordionDetails>hello</AccordionDetails>;
+
+      const wrapper = mount(<Accordion>{child}</Accordion>);
+
+      expect(wrapper.find("AccordionDetails").props().children).toEqual(
+        "hello"
+      );
+    });
+  });
```
jordanell, line 59: I don't think this test is right. We aren't testing something is being rendered. We are essentially doing:

1. Give AccordionDetails a child
2. Test did we give according details a child

I think a better test could be "does wrapped contain the child node".

packages/riipen-ui/src/components/Accordion.test.jsx
```diff
@@ -0,0 +1,150 @@
...
+        <Accordion defaultExpanded={defaultExpanded}>{child}</Accordion>
+      );
+
+      expect(wrapper.find("Collapse").props().in).toEqual(defaultExpanded);
+      expect(
+        wrapper
+          .find("Accordion")
+          .childAt(0)
+          .hasClass("expanded")
+      ).toEqual(true);
```
jordanell, line 92: This might be a good candidate for 2 separate tests. Thoughts?

packages/riipen-ui/src/components/Accordion.test.jsx
```diff
@@ -0,0 +1,150 @@
...
+
+      wrapper.find("AccordionSummary").simulate("click");
+
+      expect(wrapper.find("Collapse").props().in).toEqual(!defaultExpanded);
+      expect(
+        wrapper
+          .find("Accordion")
+          .childAt(0)
+          .hasClass("expanded")
+      ).toEqual(false);
```
jordanell, line 117: Maybe two separate tests again?

packages/riipen-ui/src/components/Accordion.test.jsx
```diff
@@ -0,0 +1,150 @@
...
+          .hasClass("expanded")
+      ).toEqual(true);
+    });
+  });
+
+  describe("expanded state", () => {
+    it("sets correct expanded class when summary is clicked", () => {
+      const defaultExpanded = true;
+      const summary = <AccordionSummary>summary</AccordionSummary>;
+      const child = <AccordionDetails>hello</AccordionDetails>;
```
jordanell, line 100: Not sure these need to be in variables. Could just in-line them.

packages/riipen-ui/src/components/Accordion.test.jsx
```diff
@@ -0,0 +1,150 @@
...
+          .childAt(0)
+          .hasClass(classes[0])
+      ).toEqual(true);
+    });
+  });
+
+  describe("defaultExpanded prop", () => {
+    it("sets expanded class with given defaultExpanded", () => {
+      const defaultExpanded = true;
+      const child = <AccordionDetails>hello</AccordionDetails>;
```
jordanell, line 80: This doesn't really need to be in a variable.

packages/riipen-ui/src/components/Accordion.test.jsx
```diff
@@ -0,0 +1,150 @@
...
+          .childAt(0)
+          .hasClass("expanded")
+      ).toEqual(false);
+    });
+  });
+
+  describe("disabled prop", () => {
+    it("applies disabled class name when disabled is true", () => {
+      const disabled = true;
+      const child = <AccordionDetails>hello</AccordionDetails>;
```
jordanell, line 124: Again, probably doesn't need to be a variable.

packages/riipen-ui/src/components/Accordion.test.jsx
```diff
@@ -0,0 +1,150 @@
...
+        wrapper
+          .find("Accordion")
+          .childAt(0)
+          .hasClass("disabled")
+      ).toEqual(true);
+    });
+
+    it("applies disabled class name when disabled is false", () => {
+      const disabled = false;
+      const child = <AccordionDetails>hello</AccordionDetails>;
```
jordanell, line 138: Same here with variable.

jordanell (changes requested):

jordanell (approved):

imogenem (approved):

PR #148, opened 2021-02-11
[PLT-6623] Tests for AccordionSummary

imogenem:
## Description
Tests for AccordionSummary

## Notes
Warning for onFocusVisible appears but all tests pass. Once Accordion is merged with master, should be able to run without the warning.

## Where to Start
AccordionSummary.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/AccordionSummary.test.jsx
+++ b/packages/riipen-ui/src/components/AccordionSummary.test.jsx
@@ -83,5 +83,18 @@ describe("<AccordionSummary>", () => {
 
       expect(wrapper.find("ButtonIcon").props().size).toEqual(iconProps.size);
     });
+
+    it("passes iconProp classes to icon classes", () => {
+      const icon = Icon;
+      const iconProps = { classes: ["small"] };
+
+      const wrapper = mount(
+        <AccordionSummary icon={icon} iconProps={iconProps} />
+      );
+
+      expect(wrapper.find("ButtonIcon").props().classes[0]).toContain(
+        iconProps.classes
+      );
+    });
   });
 });
--- a/packages/riipen-ui/src/components/AccordionSummary.test.jsx
+++ b/packages/riipen-ui/src/components/AccordionSummary.test.jsx
@@ -66,7 +66,9 @@ describe("<AccordionSummary>", () => {
 
       const wrapper = mount(<AccordionSummary icon={icon} />);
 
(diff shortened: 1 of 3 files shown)


packages/riipen-ui/src/components/AccordionSummary.test.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+  describe("icon prop", () => {
+    it("renders icon given valid custom icon prop", () => {
+      const icon = Icon;
+
+      const wrapper = mount(<AccordionSummary icon={icon} />);
+
+      expect(wrapper.find(".icon").type()).toEqual("div");
+    });
+  });
```
licarm, line 71: You can assert that the provided icon is being rendered using the assertion .containsMatchingElement() or .contains().

licarm (changes requested):
Nice!! Just one thing.

packages/riipen-ui/src/components/AccordionSummary.test.jsx
```diff
@@ -0,0 +1,87 @@
...
+
+      const wrapper = mount(<AccordionSummary icon={icon} />);
+
+      expect(
+        wrapper.find("AccordionSummary").containsMatchingElement(icon)
+      ).toEqual(true);
+    });
+  });
+
+  describe("iconProps prop", () => {
```
jordanell, line 75: I think we should add another test here. The `iconProps` prop also looks like it passes `iconProps.classes` to the icon `classes` here https://github.com/riipen/ui/blob/eed21d980e392b67816338eb5cfc816c5e877746/packages/riipen-ui/src/components/AccordionSummary.jsx#L94. 

So let's add a test to make sure that happens.

jordanell (changes requested):
Nice job!

jordanell (approved):

packages/riipen-ui/src/components/AccordionSummary.test.jsx
```diff
@@ -0,0 +1,100 @@
...
+      const icon = Icon;
+      const iconProps = { classes: ["small"] };
+
+      const wrapper = mount(
+        <AccordionSummary icon={icon} iconProps={iconProps} />
+      );
+
+      expect(wrapper.find("ButtonIcon").props().classes[0]).toContain(
+        iconProps.classes
+      );
```
licarm, line 97: Look great!

licarm (approved):

PR #149, opened 2021-02-11
[PLT-6631] tests for ButtonIcon component

licarm:
## Description

- outline prop removed from ButtonIcon because it doesn't do anything anymore
- tests in ButtonIcon.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/ButtonIcon.jsx
+++ b/packages/riipen-ui/src/components/ButtonIcon.jsx
@@ -12,7 +12,6 @@ const ButtonIcon = props => {
     color,
     component,
     disabled,
-    outline,
     size,
     theme,
     ...other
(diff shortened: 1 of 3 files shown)


packages/riipen-ui/src/components/ButtonIcon.test.jsx
```diff
@@ -0,0 +1,237 @@
...
+
+      expect(
+        wrapper
+          .find("ButtonIcon")
+          .childAt(0)
+          .hasClass("focusVisible")
+      ).toEqual(true);
+    });
+
+    it("ddoes not apply focusVisible class when blur event occurs", () => {
```
imogenem, line 223: spelling error for "does"

packages/riipen-ui/src/components/ButtonIcon.test.jsx
```diff
@@ -0,0 +1,237 @@
...
+
+      expect(
+        wrapper
+          .find("ButtonIcon")
+          .childAt(0)
+          .name()
+      ).toEqual(component);
+    });
+
+    it("sets component with 'button' when component prop is not provided", () => {
```
imogenem, line 102: Could this be rephrased to something like...

"sets component type to button when component prop is not provided"

imogenem (changes requested):

packages/riipen-ui/src/components/ButtonIcon.jsx
```diff
@@ -58,116 +57,116 @@ const ButtonIcon = props => {
...
-        `}</style>
+        .root {
+          background-color: transparent;
+          border: 0;
+          color: ${theme.palette.text.secondary};
+          cursor: pointer;
+          display: inline-flex;
+          padding: ${theme.spacing(2)}px;
+          position: relative;
+          outline: 0;
```
jordanell, line 68: Sick.

packages/riipen-ui/src/components/ButtonIcon.test.jsx
```diff
@@ -0,0 +1,237 @@
...
+      const errors = jest.spyOn(console, "error").mockImplementation();
+      const component = "abcd";
+
+      mount(<ButtonIcon component={component}>children</ButtonIcon>);
+
+      expect(errors).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe("disabled prop", () => {
```
jordanell, line 136: I think we are also missing the test "sets disabled on button when disabled prop is true".

jordanell (changes requested):

packages/riipen-ui/src/components/ButtonIcon.test.jsx
licarm: Is that the same thing as "applies disabled class when disabled prop is true"? It supposed to test line 52 in ButtonIcon.

jordanell: Sorry, I don't think I was clear.

    The test you have in this file on line 137 tests line 35 in the source file
    We should add a test to test line 53 in the source file.

jordanell (approved):

PR #150, opened 2021-02-16
[PLT-6632] tests for Checkbox component

licarm:
## Description
tests for Checkbox Component


packages/riipen-ui/src/components/checkbox.test.jsx
```diff
@@ -0,0 +1,227 @@
...
+    });
+  });
+
+  describe("color prop", () => {
+    it("sets color class name with a valid color", () => {
+      const color = "secondary";
+
+      const wrapper = mount(<Checkbox color={color} readOnly />);
+
+      expect(toJson(wrapper)).toMatchSnapshot();
```
imogenem, line 73: Remove snapshot test from color prop

packages/riipen-ui/src/components/checkbox.test.jsx
```diff
@@ -0,0 +1,227 @@
...
+        <Checkbox required={required} label={label} readOnly />
+      );
+
+      expect(
+        wrapper
+          .find("Checkbox")
+          .find("label")
+          .text()
+          .substr(-1)
+      ).toEqual("*");
```
imogenem, line 157: Nice!!

packages/riipen-ui/src/components/checkbox.test.jsx
```diff
@@ -0,0 +1,227 @@
...
+      const warning = <span>warning!</span>;
+
+      const wrapper = mount(<Checkbox warning={warning} readOnly />);
+
+      expect(wrapper.childAt(0).contains(warning)).toEqual(true);
+    });
+  });
+
+  describe("focusVisible state", () => {
+    it("does not apply focusVisible class when component is rendered", () => {
```
imogenem, line 189: Is this test necessary? I'm wondering if we only need to test after events occur.

imogenem (changes requested):
Looks great! @jordanell can you comment on the focusVisible tests?

packages/riipen-ui/src/components/checkbox.test.jsx
jordanell: Ya I think we can safely remove this. Imo makes a good point.

jordanell (changes requested):
Same comments as Imo :).

jordanell (approved):

PR #151, opened 2021-02-17
[Plt-6634] tests for ClickAway component

licarm:
## Description
- tests for ClickAway component
- checking for null children before cloneElement

### DIFF
--- a/packages/riipen-ui/src/components/ClickAway.jsx
+++ b/packages/riipen-ui/src/components/ClickAway.jsx
@@ -40,9 +40,11 @@ class ClickAway extends React.Component {
 
     const theme = this.context;
 
-    const childrenWithProps = React.Children.map(children, child =>
-      React.cloneElement(child, { onClick: this.handleClick })
-    );
+    const childrenWithProps = React.Children.map(children, child => {
+      if (!child) return null;
+
+      return React.cloneElement(child, { onClick: this.handleClick });
+    });
 
     return (
       <React.Fragment>
--- a/packages/riipen-ui/src/components/ClickAway.test.jsx
+++ b/packages/riipen-ui/src/components/ClickAway.test.jsx
@@ -1,0 +1,114 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import ClickAway from "./ClickAway";
+
+describe("<ClickAway>", () => {
+  it("renders without errors", () => {
+    let error;
+    const onClickAway = jest.fn();
+
+    try {
+      mount(<ClickAway onClickAway={onClickAway} />);
+    } catch (e) {
+      error = e;
+    }
+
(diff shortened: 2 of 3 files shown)


packages/riipen-ui/src/components/ClickAway.test.jsx
```diff
@@ -0,0 +1,114 @@
...
+  describe("onClickAway prop", () => {
+    it("sets onClick prop of children node", () => {
+      const child = <span>cheesecakes</span>;
+      const onClickAway = jest.fn();
+
+      const wrapper = mount(
+        <ClickAway onClickAway={onClickAway}>{child}</ClickAway>
+      );
+
+      expect(
```
imogenem, line 58: Can this test be simplified to something like....

```
     expect(
        wrapper
          .find("ClickAway")
          .find("span")
          .props().onClick
      ).toBeDefined();
```

Just to make it easier to read, and in line with how the next test is set up.

imogenem (commented):
Really nice work with the onClickAway prop

packages/riipen-ui/src/components/ClickAway.jsx
```diff
@@ -40,9 +40,11 @@ class ClickAway extends React.Component {
 
     const theme = this.context;
 
-    const childrenWithProps = React.Children.map(children, child =>
-      React.cloneElement(child, { onClick: this.handleClick })
-    );
+    const childrenWithProps = React.Children.map(children, child => {
+      if (!child) return null;
```
jordanell, line 44: Nice!

packages/riipen-ui/src/components/ClickAway.test.jsx
```diff
@@ -0,0 +1,114 @@
...
+      const childrenNodes = wrapper
+        .find("ClickAway")
+        .childAt(0)
+        .children();
+      expect(childrenNodes.at(0).containsMatchingElement(child)).toEqual(true);
+    });
+  });
+
+  describe("onClickAway prop", () => {
+    it("sets onClick prop of children node", () => {
```
jordanell, line 50: I wonder if this should actually be part of the children prop tests up above since we are testing children and not the actual `onClickAway` prop.

packages/riipen-ui/src/components/ClickAway.test.jsx
```diff
@@ -0,0 +1,114 @@
...
+        wrapper
+          .find("ClickAway")
+          .childAt(0)
+          .children()
+          .at(0)
+          .props().onClick
+      ).toBeDefined();
+    });
+
+    it("sets onClick prop of div element", () => {
```
jordanell, line 68: This test isn't really testing anything to do with `onClickAway` prop. I think we can actually remove this test entirely because it is actually covered by the test on line 81.

jordanell (changes requested):

jordanell (approved):

PR #152, opened 2021-02-18
[PLT-6649] tests for ListItem component

licarm:
## Description
tests for ListItem component

### DIFF
--- a/packages/riipen-ui/src/components/ListItem.test.jsx
+++ b/packages/riipen-ui/src/components/ListItem.test.jsx
@@ -64,19 +64,6 @@ describe("<ListItem>", () => {
   });
 
   describe("color prop", () => {
-    it("applies valid custom color", () => {
-      const color = "secondary";
-
-      const wrapper = mount(<ListItem color={color} />);
-
-      expect(
-        wrapper
-          .find("ListItem")
-          .childAt(0)
-          .hasClass(color)
-      ).toEqual(true);
-    });
-
     it("gives an error when given an invalid color", () => {
       const errors = jest.spyOn(console, "error").mockImplementation();
 
--- a/packages/riipen-ui/src/components/ListItem.test.jsx
+++ b/packages/riipen-ui/src/components/ListItem.test.jsx
@@ -1,0 +1,193 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import ListItem from "./ListItem";
+
+describe("<ListItem>", () => {
+  it("renders without errors", () => {
+    let error;
+
+    try {
+      mount(<ListItem />);
+    } catch (e) {
+      error = e;
+    }
+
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/ListItem.test.jsx
```diff
@@ -0,0 +1,193 @@
...
+        .childAt(0)
+        .simulate("keydown", { key: "a" });
+
+      expect(onClick).toHaveBeenCalledTimes(0);
+    });
+  });
+
+  describe("focusVisible state", () => {
+    it("does not apply focusVisible class when component is rendered", () => {
+      const wrapper = mount(<ListItem />);
```
imogenem, line 156: Same as with Checkbox, is this test needed?

imogenem (commented):
Just one comment similar to Checkbox, looks great otherwise.

packages/riipen-ui/src/components/ListItem.test.jsx
```diff
@@ -0,0 +1,182 @@
...
+    it("applies valid custom color", () => {
+      const color = "secondary";
+
+      const wrapper = mount(<ListItem color={color} />);
+
+      expect(
+        wrapper
+          .find("ListItem")
+          .childAt(0)
+          .hasClass(color)
```
jordanell, line 76: Am I crazy, or does the color class actually do nothing. Check out the component source code. There isn't actually any `.primary` or `.secondary` CSS classes. It seems like all it is really doing is affecting the `focusVisible` color. 

I think we should update this test to actually test that , or maybe move this test to be part of the focus tests down below.

packages/riipen-ui/src/components/ListItem.test.jsx
```diff
@@ -0,0 +1,182 @@
...
+
+    it("invokes onClick handler on enter keydown", () => {
+      const onClick = jest.fn();
+
+      const wrapper = mount(<ListItem onClick={onClick} />);
+
+      wrapper
+        .find("ListItem")
+        .childAt(0)
+        .simulate("keydown", { key: "Enter" });
```
jordanell, line 135: Nice!

jordanell (changes requested):

packages/riipen-ui/src/components/ListItem.test.jsx
licarm: Since the CSS uses selectors (.focusVisible), is there no way to test this prop? Because I can't move it to inline style?

jordanell: Ya that is super annoying. I always hate saying "this code is untestable" because usually that's a sign of bad code, but I also cannot think of any way to make this testable. So let's just maybe remove this test since it's not actually testing anything.

jordanell (changes requested):

jordanell (approved):

PR #153, opened 2021-02-18
[PLT-6644] Tests for Input

imogenem:
## Description
Tests for Input

## Where to Start
Input.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/Input.test.jsx
+++ b/packages/riipen-ui/src/components/Input.test.jsx
@@ -1,0 +1,206 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import Input from "./Input";
+import Typography from "./Typography";
+import InputLabel from "./InputLabel";
+
+describe("<Input>", () => {
+  it("renders without errors", () => {
+    let error;
+
+    try {
+      mount(<Input />);
+    } catch (theError) {
+      error = theError;
+    }
+
+    expect(error).toEqual(undefined);
+  });
+
+  it("renders correct snapshot", () => {
+    const wrapper = mount(<Input />);
+
+    expect(toJson(wrapper)).toMatchSnapshot();
+  });
+
+  describe("default props", () => {
+    it("sets correct default props", () => {
+      const wrapper = mount(<Input />);
+
+      expect(wrapper.find("Input").props().disabled).toEqual(false);
+      expect(wrapper.find("Input").props().multiline).toEqual(false);
+    });
+  });
+
+  describe("classes prop", () => {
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/Input.test.jsx
```diff
@@ -0,0 +1,323 @@
...
+
+      const wrapper = mount(<Input label={label} />);
+
+      expect(
+        wrapper
+          .find("Input")
+          .childAt(0)
+          .childAt(0)
+          .type()
+      ).toEqual(InputLabel);
```
licarm, line 146: I think the expect statement checks that it "renders InputLabel given label" instead of setting a label.

packages/riipen-ui/src/components/Input.test.jsx
```diff
@@ -0,0 +1,323 @@
...
+    it("sets valid custom required", () => {
+      const label = "Test";
+      const required = true;
+
+      const wrapper = mount(<Input label={label} required={required} />);
+
+      expect(wrapper.find("InputLabel").props().required).toEqual(required);
+    });
+
+    it("displays an asterisk at the end of the label when required is true", () => {
```
licarm, line 229: I think this test might be actually testing the InputLabel component. The Input component is responsible for passing it into InputLabel and InputLabel does the asterisk.

packages/riipen-ui/src/components/Input.test.jsx
```diff
@@ -0,0 +1,323 @@
...
+      expect(
+        wrapper
+          .find("InputLabel")
+          .find("label")
+          .text()
+          .substr(-1)
+      ).toEqual("*");
+    });
+
+    it("does not display an asterisk when required is false", () => {
```
licarm, line 244: Same as ^

licarm (changes requested):
Very nice!!

packages/riipen-ui/src/components/Input.test.jsx
jordanell: Ya I think we can actually change this test to check if `InputLabel` gets the prop of `children` which is set to `"Test"`.

packages/riipen-ui/src/components/Input.test.jsx
jordanell: Ya @licarm is right, so I think we can actually remove this test.

packages/riipen-ui/src/components/Input.test.jsx
jordanell: Samesies.

jordanell (changes requested):

jordanell (approved):

PR #154, opened 2021-02-18
[PLT-6647] Tests for Link component

licarm:
## Description

- Tests for Link component
- the underline prop is not testable :(

### DIFF
--- a/packages/riipen-ui/src/components/Link.test.jsx
+++ b/packages/riipen-ui/src/components/Link.test.jsx
@@ -53,7 +53,7 @@ describe("<Link>", () => {
 
       const wrapper = mount(<Link href="google.com">{child}</Link>);
 
-      expect(wrapper.contains(child)).toEqual(true);
+      expect(wrapper.find("Link").contains(child)).toEqual(true);
     });
   });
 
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/Link.test.jsx
```diff
@@ -0,0 +1,178 @@
...
+    });
+  });
+
+  describe("children prop", () => {
+    it("renders given children", () => {
+      const child = <span>link</span>;
+
+      const wrapper = mount(<Link href="google.com">{child}</Link>);
+
+      expect(wrapper.childAt(0).contains(child)).toEqual(true);
```
imogenem, line 56: Can this test use find("Link") instead of childAt(0) to make it a bit easier to read? (and similar to the other tests)

packages/riipen-ui/src/components/Link.test.jsx
```diff
@@ -0,0 +1,178 @@
...
+
+      expect(
+        wrapper
+          .find("Link")
+          .childAt(0)
+          .hasClass("focusVisible")
+      ).toEqual(false);
+    });
+  });
+});
```
imogenem, line 178: Should there be tests for the underline prop? It looks like we could move the needed info to inline style arguments.

imogenem (changes requested):

packages/riipen-ui/src/components/Link.test.jsx
licarm: underline prop is not testable :(

jordanell (approved):

PR #155, opened 2021-02-20
[PLT-6652] tests for MenuList component

licarm:
## Description
tests for MenuList component

## Notes
- keydown events simulated by enzyme won't work with `window.addEventListener()` because enzyme tests React elements, so enzyme's `simulate('keydown')` will not be detected by the listener.


packages/riipen-ui/src/components/MenuList.test.jsx
```diff
@@ -0,0 +1,299 @@
...
+
+      wrapper
+        .find("MenuList")
+        .childAt(0)
+        .simulate("click");
+
+      expect(selectChange).toHaveBeenLastCalledWith(null, expect.any(Object));
+    });
+
+    describe("when variant is selection", () => {
```
imogenem, line 182: It looks like this describe block could be removed and the test could be moved to variant. The test name for this test and the test at line 262 sound very similar.

packages/riipen-ui/src/components/MenuList.test.jsx
```diff
@@ -0,0 +1,299 @@
...
+      const selectChange = jest.fn();
+
+      mount(
+        <MenuList autoFocus={autoFocus} selectChange={selectChange}>
+          <MenuItem>one</MenuItem>
+          <MenuItem>two</MenuItem>
+        </MenuList>
+      );
+
+      expect(selectChange).toHaveBeenLastCalledWith(0, undefined);
```
imogenem, line 65: Would this test pass with just one argument, the zero? 

`toHaveBeenLastCalledWith(0)`

imogenem (changes requested):
This is a tough file to test, so great job. I'm wondering what the code coverage will look like for this file, and if more things needs to be/can be tested, like keyDownHandler().

packages/riipen-ui/src/components/MenuList.jsx
```diff
@@ -10,7 +10,7 @@ class MenuList extends React.Component {
 
   static propTypes = {
     /**
-     * Wether to select the first element on list open
+     * Whether to select the first element on list open
```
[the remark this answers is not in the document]
jordanell: 10 points for Gryffindor

packages/riipen-ui/src/components/MenuList.test.jsx
```diff
@@ -0,0 +1,299 @@
...
+      const component = wrapper.find("MenuList");
+      expect(component.props().selectedIndex).toEqual(
+        defaultProps.selectedIndex
+      );
+      expect(component.props().variant).toEqual(defaultProps.variant);
+    });
+  });
+
+  describe("autoFocus prop", () => {
+    it("selects first element on list open with true autoFocus and default selectedIndex", () => {
```
jordanell, line 54: This sentence is sort of confusing to me. I think what we are testing is that we auto focus the first element of the list on mount, so what if we said "selects first item of list with ..."

packages/riipen-ui/src/components/MenuList.test.jsx
```diff
@@ -0,0 +1,299 @@
...
+          .at(0)
+          .childAt(0)
+          .simulate("click");
+
+        expect(selectChange).not.toHaveBeenCalled();
+      });
+    });
+  });
+
+  describe("selectedIndex prop", () => {
```
jordanell, line 223: Should we add a test for selectedIndex + autoFocus? I don't see this combination.

packages/riipen-ui/src/components/MenuList.test.jsx
jordanell: Ya I am also confused about this describe block. Looking at the two tests within it, I wonder if they are better off being part of the children tests, or even remove the describe block entirely and just put those two tests at the same level of the other selectChange test. Up to you.

jordanell (changes requested):
Nice. I think we are missing some tests for the `keyDownHandler` code as part of the component though.

packages/riipen-ui/src/components/MenuList.test.jsx
licarm: I like them being part of children tests

packages/riipen-ui/src/components/MenuList.test.jsx
licarm: no :(

packages/riipen-ui/src/components/MenuList.test.jsx
```diff
@@ -0,0 +1,323 @@
...
+
+      expect(
+        wrapper
+          .find("MenuItem")
+          .at(0)
+          .props().selected
+      ).toBeUndefined();
+    });
+
+    it("prioritizes selectedIndex over autoFocus when autoFocus is true", () => {
```
jordanell, line 258: Nice

packages/riipen-ui/src/components/MenuList.test.jsx
```diff
@@ -0,0 +1,323 @@
...
+          <MenuItem>one</MenuItem>
+          <MenuItem>two</MenuItem>
+        </MenuList>
+      );
+
+      expect(selectChange).not.toHaveBeenCalled();
+      expect(
+        wrapper
+          .find("MenuItem")
+          .at(selectedIndex)
```
jordanell, line 279: Slick

jordanell (approved):

imogenem (approved):

PR #159, opened 2021-02-23
[PLT-6662] Tests for Table component

licarm:
## Description
Tests for Table component

## Notes
The `forwardedRef` prop sets the `ref` of the `<table>` element. `ref` itself is React implemented and does not show up when the component mounts.

### DIFF
--- a/packages/riipen-ui/src/components/Table.test.jsx
+++ b/packages/riipen-ui/src/components/Table.test.jsx
@@ -1,0 +1,129 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import defaultTheme from "../themes/default";
+
+import Table from "./Table";
+import TableHeader from "./TableHeader";
+
+describe("<Table>", () => {
+  it("renders without errors", () => {
+    let error;
+
+    try {
+      mount(<Table />);
+    } catch (e) {
+      error = e;
+    }
+
+    expect(error).toEqual(undefined);
+  });
+
+  it("renders correct snapshot", () => {
+    const wrapper = mount(<Table />);
+
+    expect(toJson(wrapper)).toMatchSnapshot();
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/Table.test.jsx
```diff
@@ -0,0 +1,129 @@
...
+      const color = "grey400";
+
+      const wrapper = mount(<Table backgroundColor={color} />);
+
+      expect(wrapper.find("JSXStyle").props().dynamic).toContain(
+        `${defaultTheme.palette.grey[400]}`
+      );
+    });
+
+    it("sets transparent color given transparent backgroundColor", () => {
```
imogenem, line 50: The grey and transparent colors are two difference oneOf types for backgroundColor, so it doesn't seem like this second test is necessary.

imogenem (changes requested):

packages/riipen-ui/src/components/Table.test.jsx
```diff
@@ -0,0 +1,129 @@
...
+
+  describe("backgroundColor prop", () => {
+    it("sets custom valid backgroundColor with mapped hex color", () => {
+      const color = "grey400";
+
+      const wrapper = mount(<Table backgroundColor={color} />);
+
+      expect(wrapper.find("JSXStyle").props().dynamic).toContain(
+        `${defaultTheme.palette.grey[400]}`
+      );
```
jordanell, line 47: Nice

packages/riipen-ui/src/components/Table.test.jsx
jordanell: Ya I agree. We are testing the same thing twice here.

jordanell (changes requested):

jordanell (approved):

PR #160, opened 2021-02-24
[PLT-6650] Tests for Menu

imogenem:
## Description
Tests for Menu

## Where to Start
Menu.test.jsx


packages/riipen-ui/src/components/Menu.test.jsx
```diff
@@ -0,0 +1,333 @@
...
+        .find("MenuItem")
+        .childAt(0)
+        .simulate("click");
+
+      expect(onChange).toHaveBeenCalledWith(null, expect.any(Object));
+    });
+  });
+
+  describe("onClose prop", () => {
+    it("invokes onClose when child is clicked", () => {
```
licarm, line 253: Can we check the case when `closeOnClick` is false?

packages/riipen-ui/src/components/Menu.test.jsx
```diff
@@ -0,0 +1,333 @@
...
+
+      const wrapper = mount(<Menu marginThreshold={marginThreshold} />);
+
+      expect(wrapper.find("Popover").props().marginThreshold).toEqual(
+        marginThreshold
+      );
+    });
+  });
+
+  describe("onChange prop", () => {
```
licarm, line 228: Maybe we can test that it doesn't break when onChange prop is not provided.

packages/riipen-ui/src/components/Menu.test.jsx
```diff
@@ -0,0 +1,333 @@
...
+      wrapper
+        .find("MenuItem")
+        .childAt(0)
+        .simulate("click");
+
+      expect(onChange).toHaveBeenCalledWith(null, expect.any(Object));
+    });
+  });
+
+  describe("onClose prop", () => {
```
licarm, line 252: Or when onClose prop isn't provided.

licarm (changes requested):
Great job! Wondering if a few tests can be added.

packages/riipen-ui/src/components/Menu.test.jsx
```diff
@@ -0,0 +1,333 @@
...
+    });
+  });
+
+  describe("children prop", () => {
+    it("renders given children", () => {
+      const child = <div>child</div>;
+      const anchorEl = ReactTestUtils.renderIntoDocument(
+        <div>
+          <AppBar />
+        </div>
```
jordanell, line 99: Do you need the wrapping `<div>`'s here? Seems like in some places you use them and others you don't.

jordanell (approved):

packages/riipen-ui/src/components/Menu.test.jsx
imogenem: The wrapping divs allow for Popover's children to be rendered. So I think they are needed when testing props passed to MenuList. 

The wrapping divs had to be left out for the anchorEl test at line 53, otherwise I would also have to pass in the required children for MenuList.

packages/riipen-ui/src/components/Menu.test.jsx
imogenem: ^ what kind of break are you referring to? The snapshot test passes without this prop

packages/riipen-ui/src/components/Menu.test.jsx
jordanell: Ok cool, that's good to know!

packages/riipen-ui/src/components/Menu.test.jsx
licarm: Basically to test the handleChange function, line 124 makes sure that if onChange prop isn't provided it won't call an undefined function.

licarm (approved):

jordanell (approved):

PR #162, opened 2021-02-25
[PLT-6635] Tests for Collapse

imogenem:
## Description
Tests for Collapse

## Where to Start
Collapse.test.jsx


packages/riipen-ui/src/components/Collapse.test.jsx
```diff
@@ -0,0 +1,114 @@
...
+          .find("Collapse")
+          .childAt(0)
+          .childAt(0)
+          .type()
+      ).toEqual(component);
+    });
+  });
+
+  describe("in prop", () => {
+    it("sets state to entered when in prop is true", () => {
```
licarm, line 86: Is `entered` a state? It might be good to rename the test to something like "applies entered class..." It's also consistent with the expect using `hasClass("entered")`.

packages/riipen-ui/src/components/Collapse.test.jsx
```diff
@@ -0,0 +1,114 @@
...
+      expect(
+        wrapper
+          .find("Collapse")
+          .childAt(0)
+          .childAt(0)
+          .hasClass("entered")
+      ).toEqual(true);
+    });
+
+    it("sets state to exited when in prop is false", () => {
```
licarm, line 100: Same thing for exited.

licarm (changes requested):
Nice job! just two things.

jordanell (approved):
Agreed with Carmen's comments.

jordanell (approved):

licarm (approved):

PR #164, opened 2021-03-01
[PLT-6639] Tests for Form

imogenem:
## Description
Tests for Form

## Notes
Tried testing the errorScroll and enter props with mock & spy functions, but could not find a test that worked with functional components.

## Where to Start
Form.test.jsx

### DIFF
--- a/packages/riipen-ui/src/components/Form.test.jsx
+++ b/packages/riipen-ui/src/components/Form.test.jsx
@@ -208,14 +208,6 @@ describe("<Form>", () => {
       ).toEqual(errors[1]);
     });
 
-    it("renders errors within Typography element if errors are strings", () => {
-      const errors = ["Error", "Sad"];
-
-      const wrapper = mount(<Form errors={errors} />);
-
-      expect(wrapper.find("Typography").contains(errors[0])).toEqual(true);
-    });
-
     it("renders errors from given object", () => {
       const errors = {
         Error: "Please provide more cheese",
--- a/packages/riipen-ui/src/components/Form.test.jsx
+++ b/packages/riipen-ui/src/components/Form.test.jsx
@@ -1,0 +1,161 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import Form from "./Form";
+import Typography from "./Typography";
+
+describe("<Form>", () => {
+  it("renders without errors", () => {
+    let error;
+
+    try {
+      mount(<Form />);
+    } catch (theError) {
+      error = theError;
+    }
+
+    expect(error).toEqual(undefined);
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/Form.test.jsx
```diff
@@ -0,0 +1,161 @@
...
+    it("appends higher order values to classes prop with withClass decorator", () => {
+      const classes = ["riipen", "riipen-form"];
+
+      const wrapper = mount(<Form />);
+
+      expect(wrapper.find("Form").props().classes).toEqual(classes.sort());
+    });
+  });
+
+  describe("error prop", () => {
```
licarm, line 69: Can we check that the error gets displayed using `.containsMatchingElement()` or something similar? And maybe be check that it gets put inside a `Typography` element if it is a string.

packages/riipen-ui/src/components/Form.test.jsx
```diff
@@ -0,0 +1,161 @@
...
+      expect(
+        wrapper
+          .find("Form")
+          .childAt(0)
+          .hasClass("errorContainer")
+      ).toEqual(true);
+    });
+  });
+
+  describe("errors prop", () => {
```
licarm, line 107: Same as above. Also wondering if we can check that `renderErrorsObject` processes errors correctly when given as an object, or object inside an object.

licarm (changes requested):

packages/riipen-ui/src/components/Form.test.jsx
imogenem: Added both plus a test for `renderErrorNode`

packages/riipen-ui/src/components/Form.test.jsx
```diff
@@ -0,0 +1,247 @@
...
+      ).toEqual(errors[1]);
+    });
+
+    it("renders errors within Typography element if errors are strings", () => {
+      const errors = ["Error", "Sad"];
+
+      const wrapper = mount(<Form errors={errors} />);
+
+      expect(wrapper.find("Typography").contains(errors[0])).toEqual(true);
+    });
```
jordanell, line 217: I feel like these are testing the same thing? Do we need both tests?

jordanell (changes requested):
Just one simplification. Looks great!

packages/riipen-ui/src/components/Form.test.jsx
imogenem: I'll take it out, it's meant to test that an error is rendered within a Typography element, but that's already covered in the error prop block.

jordanell (approved):

licarm (approved):

PR #166, opened 2021-03-04
Fix build and update dependiecies

stredger:
## Description

Seems like upgrading to npm 7 and re running the install causes a few issues. I did a few things here to get things working again.
1. Upgraded react-docgen
2. Added a check to skip test files / snapshots when building for react-docgen
3. Updated the jest config to look in the src dir so it doesnt pick up files from a build
4. Removed some dependency warnings
5. Updated (and fixed) the required npm engine in both packages

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,8 +79,8 @@ jobs:
             sudo npm install -g npm
       - restore_cache:
           keys:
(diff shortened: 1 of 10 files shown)


jordanell (approved):
Thanks for the support on this.

imogenem (approved):
Thanks for your help!

PR #167, opened 2021-03-04
[PLT-7071] Paper Component

imogenem:
## Description
Added Paper component code, tests, and docs.

## Where to Start
Paper.jsx
Paper.test.jsx

### DIFF
--- a/packages/riipen-ui-docs/src/pages/components/paper/Elevation.jsx
+++ b/packages/riipen-ui-docs/src/pages/components/paper/Elevation.jsx
@@ -4,7 +4,7 @@ import Paper from "riipen-ui/components/Paper";
 
 export default function Elevation() {
   const style = {
-    marginRight: "10px"
+    marginRight: "20px"
   };
 
   return (
--- a/packages/riipen-ui-docs/src/pages/components/paper/Elevation.jsx
+++ b/packages/riipen-ui-docs/src/pages/components/paper/Elevation.jsx
@@ -1,0 +1,40 @@
+import React from "react";
+
+import Paper from "riipen-ui/components/Paper";
+
+export default function Elevation() {
+  const style = {
+    marginRight: "10px"
+  };
+
+  return (
+    <React.Fragment>
+      <Paper
+        style={{
+          height: "100px",
+          width: "100px"
+        }}
+      />
+
+      <span style={style} />
+
+      <Paper
+        elevation={1}
+        style={{
+          height: "100px",
+          width: "100px"
+        }}
+      />
+
+      <span style={style} />
+
+      <Paper
+        elevation={10}
+        style={{
+          height: "100px",
+          width: "100px"
+        }}
+      />
+    </React.Fragment>
+  );
+}
--- a/packages/riipen-ui-docs/src/pages/components/paper/Outline.jsx
+++ b/packages/riipen-ui-docs/src/pages/components/paper/Outline.jsx
@@ -4,7 +4,7 @@ import Paper from "riipen-ui/components/Paper";
 
 export default function Outline() {
   const style = {
-    marginRight: "10px"
+    marginRight: "20px"
   };
 
   return (
--- a/packages/riipen-ui-docs/src/pages/components/paper/Outline.jsx
+++ b/packages/riipen-ui-docs/src/pages/components/paper/Outline.jsx
@@ -1,0 +1,32 @@
+import React from "react";
+
+import Paper from "riipen-ui/components/Paper";
+
+export default function Outline() {
+  const style = {
+    marginRight: "10px"
+  };
+
+  return (
+    <React.Fragment>
+      <Paper
+        variant="outlined"
+        style={{
+          height: "100px",
+          width: "100px"
+        }}
+      />
+
+      <span style={style} />
+
+      <Paper
+        variant="outlined"
+        rounded={true}
+        style={{
+          height: "100px",
+          width: "100px"
+        }}
+      />
+    </React.Fragment>
+  );
+}
--- a/packages/riipen-ui-docs/src/pages/components/paper/Outline.jsx
+++ b/packages/riipen-ui-docs/src/pages/components/paper/Outline.jsx
@@ -10,18 +10,14 @@ export default function Outline() {
   return (
     <React.Fragment>
       <Paper
-        variant="outlined"
         style={{
           height: "100px",
           width: "100px"
         }}
       />
-
       <span style={style} />
-
       <Paper
         variant="outlined"
-        rounded={true}
         style={{
           height: "100px",
           width: "100px"
--- a/packages/riipen-ui-docs/src/pages/components/paper/paper.md
+++ b/packages/riipen-ui-docs/src/pages/components/paper/paper.md
@@ -1,0 +1,17 @@
+# Paper
+
+<p class="description">Paper wraps content within a contained background.</p>
+
+Paper can contain a number of elements, including Typography to display messages.
+
+## Elevation
+
+The `elevation` prop adjusts the height of the box shadow displayed behind the paper.
+
+{{"demo": "pages/components/paper/Elevation.js"}}
+
+## Outline
+
+If you need an outlined surface, use the `variant` prop. Outlines can either be square or rounded, which are determined by the `rounded` prop.
+
+{{"demo": "pages/components/paper/Outline.js"}}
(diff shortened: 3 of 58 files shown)


packages/riipen-ui-docs/src/pages/components/paper/Elevation.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+  return (
+    <React.Fragment>
+      <Paper
+        style={{
+          height: "100px",
+          width: "100px"
+        }}
+      />
+
```
jordanell, line 18: We general tend to avoid blank lines like this in our React code. I wouldn't say one way is better than the other, but we should stay consistent with the rest of the code. Go ahead and remove these blank lines in this file and the outline file.

packages/riipen-ui-docs/src/pages/components/paper/Outline.jsx
```diff
@@ -0,0 +1,32 @@
...
+      <span style={style} />
+
+      <Paper
+        variant="outlined"
+        rounded={true}
+        style={{
+          height: "100px",
+          width: "100px"
+        }}
+      />
```
jordanell, line 29: Both of these examples are showing outlined. I think in this file we should show one outlined and one not outlined, and that we should actually move the rounding / not rounding to a different file to show those off in a separate section. Thoughts? I think it would be just nice from a user standpoint to see those things separately.

packages/riipen-ui-docs/src/pages/components/paper/paper.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Elevation
+
+The `elevation` prop adjusts the height of the box shadow displayed behind the paper.
+
+{{"demo": "pages/components/paper/Elevation.js"}}
+
+## Outline
+
+If you need an outlined surface, use the `variant` prop. Outlines can either be square or rounded, which are determined by the `rounded` prop.
```
jordanell, line 15: If you take my comment above, this line will need updating as well.

packages/riipen-ui/src/components/Paper.jsx
```diff
@@ -0,0 +1,111 @@
...
+
+          .outlined {
+            border: 1px solid ${theme.palette.divider};
+        }
+            
+            .rounded {
+                border-radius: ${theme.shape.borderRadius.md};
+            }
+      `}
+      </style>
```
jordanell, line 63: You indentation is off here. It should be:

```js
<React.Fragment>
  <Component ...>
    {children}
  </Component>
  <style jsx>
    ....
  </style>
</React.Component>
```

packages/riipen-ui/src/components/Paper.jsx
```diff
@@ -0,0 +1,111 @@
...
+    rounded ? "rounded" : null,
+    variant === "outlined" ? "outlined" : "elevation"
+  );
+  
+  return (
+    <React.Fragment>
+    <Component 
+    className={className}
+      {...other}
+    >
```
jordanell, line 34: This might read nicer if it was all on one line.

packages/riipen-ui/src/components/Paper.jsx
```diff
@@ -0,0 +1,111 @@
...
+      {children}
+      </Component>
+     
+      <style jsx>
+        {`
+          .root {
+            background-color: ${theme.palette.common.white};
+            color: ${theme.palette.text.primary}
```
jordanell, line 42: I think you can actually remove this. This would define any colour for text inside the component, but we don't actually care what someone renders inside the paper, that should be left to them.

packages/riipen-ui/src/components/Paper.jsx
```diff
@@ -0,0 +1,111 @@
...
+            padding: ${theme.spacing(3)}px;
```
jordanell, line 43: Now that I see this, I actually think this should be a prop. I would maybe add:

```js
spacing: PropTypes.number,
```
to the props, then apply it here. Maybe the default prop value can be 3 as well, but I would be more tempted to move it to 2 because Riipen uses padding 2 a lot.

packages/riipen-ui/src/components/Paper.jsx
```diff
@@ -0,0 +1,111 @@
...
+            z-index: ${theme.zIndex.low};
```
jordanell, line 44: I'd remove this. `z-index` can be a pretty dangerous thing to play with. A good rule of thumb is if you don't absolutely need to set, just leave it as whatever default the browser gives it.

packages/riipen-ui/src/components/Paper.jsx
```diff
@@ -0,0 +1,111 @@
...
+  classes: PropTypes.array,
+
+  /**
+   * The component used for the root node.
+   * Either a string to use a HTML element or a component.
+   */
+  component: PropTypes.elementType,
+
+  /**
+   * The elevation of the box shadow.
```
jordanell, line 87: What about "The elevation of the box shadow with respect to the current theme." just so people know it directly corresponds to the current theme's box shadow levels.

packages/riipen-ui/src/components/Paper.jsx
```diff
@@ -0,0 +1,111 @@
...
+   */
+  elevation: PropTypes.number,
+
+  /**
+   * If `false`, rounded corners are disabled.
```
jordanell, line 92: Since `false` is the default, you might consider rewriting this to "If `true`, rounded corners are enabled."

jordanell (changes requested):
Looking good, mostly just small suggestions. One other thing I noticed was that the browser tab for the component documentation shows "paper" instead of "Paper", see below.

Screenshot from 2021-03-04 16-30-48

We will want that to be capitalized.

packages/riipen-ui-docs/src/pages/components/paper/Outline.jsx
imogenem: I thought about it too, but it's pretty hard to see the rounded corner without the outline. I've adjusted it for now so you can have a look. Maybe the color of the background can be adjusted just in `Rounded.jsx`

packages/riipen-ui/src/components/Paper.jsx
imogenem: Should the background color stay in there though?

packages/riipen-ui-docs/src/pages/components/paper/Outline.jsx
jordanell: I just pulled but don't see the update. Maybe you haven't pushed yet and I am too eager :P.

packages/riipen-ui/src/components/Paper.jsx
imogenem: I'm tempted to call the prop `padding` to be in line with the TableCell component. Grid uses `spacing` as a prop, but it has a slightly different meaning.

jordanell: Sounds good to me :).

packages/riipen-ui/src/components/Paper.jsx
```diff
@@ -0,0 +1,106 @@
...
+  return (
+    <React.Fragment>
+      <Component className={className} {...other} >
+        {children}
+      </Component>
+      <style jsx>{`
+        .root {
+          background-color: ${theme.palette.common.white};
+          padding: ${theme.spacing[padding]};
+          display: inline-block;
```
jordanell, line 39: Can we make these in alphabetical order? We like to keep properties alphabetical but unfortunately there is not way to automate this with `styled-jsx` :( so we have to do it manually in the code review process.

packages/riipen-ui/src/components/Paper.jsx
```diff
@@ -0,0 +1,106 @@
...
+   */
+  component: PropTypes.elementType,
+
+  /**
+   * The elevation of the box shadow with respect to the current theme.
+   */
+  elevation: PropTypes.number,
+
+  /**
+   * The amount of padding the cell should have.
```
jordanell, line 81: "The amount of padding the paper should have". Paper not cell.

packages/riipen-ui/src/components/Paper.test.jsx
```diff
@@ -0,0 +1,180 @@
...
+        wrapper
+          .find("Paper")
+          .find("JSXStyle")
+          .props().dynamic
+      ).toContain(defaultTheme.shadows[elevation]);
+    });
+  });
+
+  describe("padding prop", () => {
+    it("sets padding for the component cell", () => {
```
jordanell, line 116: I would maybe rename to not use the word "cell". See comment above.

jordanell (changes requested):
Looking good! I actually like the corners docs on it's own. We can always change later :).

A few small clean up things left.

jordanell (approved):

PR #168, opened 2021-03-05
[PLT-7072] Add Switch component

licarm:
## Description
- Switch component
- Unit tests
- Component page
- Component API page

## Notes
Thanks to Imo for helping me with the `size` prop.
I followed this [tutorial](https://blog.prototypr.io/create-a-scenic-day-and-night-toggle-button-with-html-and-css-37a8e2532e68).

## Where to Start
- packages/riipen-ui/src/components/Switch.jsx
- packages/riipen-ui/src/components/Switch.test.jsx
- packages/riipen-ui-docs/src/pages/components/switch/


packages/riipen-ui/src/components/Switch.jsx
```diff
@@ -0,0 +1,219 @@
...
+    id,
+    onChange,
+    required,
+    size,
+    value
+  } = props;
+
+  const theme = React.useContext(ThemeContext);
+
+  const [checked, setChecked] = useState(checkedProp);
```
jordanell, line 23: I don't think we actually want this. You mixed up two forms of input here, both "controlled" and "uncontrolled" inputs.

https://stackoverflow.com/questions/42522515/what-are-react-controlled-components-and-uncontrolled-components

By having a `checked` prop you are enabled a controlled component, but by having this state transition you are also enabling a uncontrolled component. When build react we should never mix the two as it can have really weird side effects.

Looking at our other components, most notable `<Checkbox>`, I see we have opted for controlled components only. We should update this component to be more like the checkbox. Only have the `checked` prop and not this internal state. You will also need to update the demo pages slightly to accommodate for this. Checkout the checkbox demo pages for good examples on how to handle this.

packages/riipen-ui/src/components/Switch.test.jsx
```diff
@@ -0,0 +1,162 @@
...
+    it("changes checked state on change event", () => {
+      const checked = false;
+
+      const wrapper = mount(<Switch checked={checked} />);
+
+      wrapper.find("input").simulate("change");
+
+      expect(wrapper.find("input").props().checked).toEqual(!checked);
+    });
+  });
```
jordanell, line 57: You can remove this test as we moved to controlled components.

jordanell (changes requested):
Component looks really good! Some small changes and then one new thing you might not have seen before around controlled vs uncontrolled components.

packages/riipen-ui/src/components/Switch.jsx
```diff
@@ -0,0 +1,213 @@
...
+        onChange={onChange}
+        required={required}
+        type="checkbox"
+        value={value}
+        {...other}
+      />
+      <style jsx>{`
+        input {
+          background-color: ${theme.palette.grey[400]};
+          cursor: pointer;
```
jordanell, line 40: I think we should also not have `cursor: pointer` if `other.readOnly` is `true` since it is unclickable in that case.

jordanell (changes requested):
Just one last thing. Looking good!

packages/riipen-ui/src/components/Switch.jsx
licarm: Should we still have the shadow on hover if readOnly is true? @jordanell

jordanell (approved):

PR #170, opened 2021-03-12
[PLT 5638] Additional event functionality for Checkbox

imogenem:
## Description
Checkboxes can be set as checked when they are tabbed to and when Enter is pressed.

## Notes
Thanks Carmen for help with the event!

## Where to Start
Checkbox.jsx


jordanell (approved):
Amazing!

I think we should fix the Firefox styling issues in another PR before we deploy the new version as well.

imogenem:
> Amazing!
> 
> I think we should fix the Firefox styling issues in another PR before we deploy the new version as well.

PR #175, opened 2021-03-25
Chip Icon

jordanell:
## Description

1. Chip icon now accepts any prop type

## Notes

This is to make `<Chip>` more conducive to inversion of control.


stredger (approved):
nice

PR #176, opened 2021-03-26
[PLT-7175] Add scroll bar to Editor text area

imogenem:
## Description
Added max height and scroll bar to stop text overflow.

## Screenshots
Screen Shot 2021-03-26 at 4 19 31 PM

## Where to Start
packages/riipen-ui/src/components/Editor.jsx

### DIFF
--- a/packages/riipen-ui-docs/src/pages/components/rich-text-editor/Usage.jsx
+++ b/packages/riipen-ui-docs/src/pages/components/rich-text-editor/Usage.jsx
@@ -31,6 +31,12 @@ export default function Usage() {
       />
       <div style={style} />
       <RichTextEditor
+        id="editor-max-height"
+        label="Max Height"
+        maxHeight="200px"
+      />
+      <div style={style} />
+      <RichTextEditor
         id="editor-placeholder"
         label="Placeholder"
         placeholder="I'm holding the place!"
--- a/packages/riipen-ui/src/components/Editor.jsx
+++ b/packages/riipen-ui/src/components/Editor.jsx
@@ -115,13 +115,13 @@ class Editor extends React.Component {
     ariaLabelledBy: PropTypes.string,
 
     /**
-     * Optional array of whitelisted styles
+     * Optional array of whitelisted styles.
      * ex. ['code-block', 'BOLD', 'LINK']
      */
     controlWhitelist: PropTypes.arrayOf(PropTypes.string),
 
     /**
-     * Position of controls for the editor
+     * Position of controls for the editor.
      */
     controlPosition: PropTypes.oneOf(["top", "bottom"]),
 
@@ -131,7 +131,7 @@ class Editor extends React.Component {
     decorator: PropTypes.shape({ type: PropTypes.oneOf([CompositeDecorator]) }),
 
     /**
-     * If the error style should be shown or not
+     * If the error style should be shown or not.
      */
     error: PropTypes.any,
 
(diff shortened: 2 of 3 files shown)


packages/riipen-ui/src/components/Editor.jsx
```diff
@@ -256,8 +256,9 @@ class Editor extends React.Component {
         font-size: ${theme.typography.body1.fontSize};
 
         /* 80px - 2 * 1px borders  */
+        max-height: 300px;
```
jordanell, line 259: Can we actually make this a prop? Would be useful if we need to use this component in different places.

I would suggest a default value of just "auto".

packages/riipen-ui/src/components/Editor.jsx
```diff
@@ -256,8 +256,9 @@ class Editor extends React.Component {
...
         min-height: 78px;
-        overflow: auto;
+        overflow-y: scroll;
```
jordanell, line 261: I _think_ if you have a `max-height` you actually can leave this as `overflow-y: auto` which will only show the scroll bar if it is needed. A value of `scroll` will _always_ show the bar even if it's not needed. Can you test that and confirm?

packages/riipen-ui/src/components/Editor.jsx
imogenem: @jordanell so the prop would just be optional, and it defaults to auto height if that prop isn't provided?

jordanell: @imogenem that sounds pretty good to me!

packages/riipen-ui/src/components/Editor.jsx
imogenem: The scroll bar shows up with a given max height and `overflow-y: auto`

packages/riipen-ui-docs/src/pages/components/rich-text-editor/Usage.jsx
```diff
@@ -30,6 +30,12 @@ export default function Usage() {
         initialValue="hi!"
       />
       <div style={style} />
+      <RichTextEditor
+        id="editor-max-height"
+        label="Max Height"
+        maxHeight="200px"
+      />
```
jordanell, line 37: Good thinking.

packages/riipen-ui/src/components/Editor.jsx
```diff
@@ -256,8 +261,9 @@ class Editor extends React.Component {
         font-size: ${theme.typography.body1.fontSize};
 
         /* 80px - 2 * 1px borders  */
+        max-height: ${maxHeight || "auto"};
```
jordanell, line 264: I think instead of `|| "auto"` here, we should just make the `maxHeight` props have a default value of "auto".

packages/riipen-ui/src/components/Editor.jsx
```diff
@@ -141,15 +141,20 @@ class Editor extends React.Component {
...
     /**
-     * Function to execute when editor content changes, gets html value of editor
+     * Function to execute when editor content changes, gets html value of editor.
      */
     onChange: PropTypes.func,
 
+    /**
+     * The max height of the Editor text area.
+     */
+    maxHeight: PropTypes.number,
```
jordanell, line 156: I don't think `number` is right here. In the example above you used `"200px"` as a value, which is a string. I think we want string here instead. This will cover all cases like "200px", "75%", "5em", "auto", etc.

packages/riipen-ui/src/components/RichTextEditor.jsx
```diff
@@ -43,6 +43,11 @@ class RichTextEditor extends React.Component {
      */
     label: PropTypes.string,
 
+    /**
+     * The max height of the Editor text area.
+     */
+    maxHeight: PropTypes.number,
```
jordanell, line 49: This should be "string" as well.

jordanell (changes requested):

jordanell (approved):

PR #177, opened 2021-03-29
[PLT-7135]: Add filters, activeFiltersCount props to TableGenerator

ericauv:
## Description
We want to be able to filter tables that are created
by TableGenerator. Add a prop to render a filters component.
Also add an activeFiltersCount prop so that
we can render the number of filters that are currently active.


jordanell:
@ericauv I am not sure we actually want this code in `ui`. This seems very specific to Riipen's platform front end and not generic to the table code itself.

ericauv:
This pr is not needed, we have this functionality in `EntityTable` -- we should be wrapping Tables that we want to render within an `EntityTable` when we need `EntityTable` functionality.

PR #183, opened 2021-04-06
[PLT-7203]: Add grey color to InputLabel

ericauv:
## Description
New designs call for input labels to have #373737 color,
add grey as an option for our input labels.


daelynj (approved):
sweet!

PR #184, opened 2021-04-06
Add Underlined and Default Input variants

ksmort:
## Description
Add new `variant` prop that styles the Input component.
The `default` variant is the current styling.  The `underlined` variant displays only an underline.
Update tests and snapshots.

## Screenshots
Screenshot_2021-04-06 Input Riipen UI

## Where to Start

### DIFF
--- a/packages/riipen-ui/src/components/Input.jsx
+++ b/packages/riipen-ui/src/components/Input.jsx
@@ -74,12 +74,12 @@ const Input = ({
         {...other}
       />
       {error && (
-        <Typography color="negative" variant="body2">
+        <Typography color="negative" component="span" variant="body2">
           {error}
         </Typography>
       )}
       {warning && (
-        <Typography color="secondary" variant="body2">
+        <Typography color="secondary" component="span" variant="body2">
           {warning}
         </Typography>
       )}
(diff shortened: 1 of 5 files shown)


packages/riipen-ui/src/components/Input.jsx
```diff
@@ -48,7 +47,8 @@ const Input = props => {
     error ? "error" : null,
     disabled ? "disabled" : null,
     warning ? "warning" : null,
-    focusVisible ? "focusVisible" : null
+    focusVisible ? "focusVisible" : null,
+    variant
```
ksmort, line 51: New prop/class

packages/riipen-ui/src/components/Input.jsx
```diff
@@ -121,6 +124,12 @@ const Input = props => {
           border-color: ${theme.palette.negative.main};
         }
 
+        .underlined {
+          background-color: transparent;
```
ksmort, line 128: Wasn't sure if we should make the background colour white or transparent (see screenshot).

packages/riipen-ui/src/components/Input.jsx
```diff
@@ -121,6 +124,12 @@ const Input = props => {
...
+          border: none;
```
ksmort, line 129: Need to override default input border

packages/riipen-ui/src/components/Input.jsx
stredger: i think it looks great

stredger (approved):

PR #185, opened 2021-04-06
[PLT-7268] Paper Block

jordanell:
## Description

1. Switch the `<Paper>` component to use default CSS `display` attribute per component.

### DIFF
--- a/packages/riipen-ui/src/components/Paper.jsx
+++ b/packages/riipen-ui/src/components/Paper.jsx
@@ -35,7 +35,6 @@ const Paper = props => {
       <style jsx>{`
         .root {
           background-color: ${theme.palette.common.white};
-          display: inline-block;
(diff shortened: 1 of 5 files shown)


packages/riipen-ui/src/components/Paper.jsx
```diff
@@ -35,7 +35,6 @@ const Paper = props => {
       <style jsx>{`
         .root {
           background-color: ${theme.palette.common.white};
-          display: inline-block;
```
jordanell, line 38: This is the real fix here. Now this component will render with default `display` depending on its Component type.

daelynj: Oh yeah that's perfect

daelynj (approved):
👌

imogenem (approved):
Makes sense :)

PR #186, opened 2021-04-06
Bump version to 0.4.16

ksmort:
## Description
Bump UI version to 0.4.16

### DIFF
--- a/packages/riipen-ui/package-lock.json
+++ b/packages/riipen-ui/package-lock.json
@@ -1,11 +1,11 @@
 {
   "name": "riipen-ui",
-  "version": "0.4.15",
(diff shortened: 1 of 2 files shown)


jordanell (commented):
We don't want to do this manually. Instead do:

```bash
$ cd packages/riipen-ui
$ npm run release
```

Then just press "y" when it asks if you want to make a patch.

ksmort:
> We don't want to do this manually. Instead do:
> 
> ```shell
> $ cd packages/riipen-ui
> $ npm run release
> ```
> 
> Then just press "y" when it asks if you want to make a patch.

PR #187, opened 2021-04-07
Add Input size prop

ksmort:
## Description
Add `size` prop to Input component.  This prop changes the padding inside the input.
Update tests and docs.

## Screenshots
Screenshot_2021-04-07 Input Riipen UI(2)

## Where to Start

### DIFF
--- a/packages/riipen-ui-docs/src/pages/components/input/Variant.jsx
+++ b/packages/riipen-ui-docs/src/pages/components/input/Variant.jsx
@@ -6,9 +6,9 @@ import Input from "riipen-ui/components/Input";
 export default function Variant() {
   return (
     <Form>
-      <Input id="name" label="Biography" variant="default" />
+      <Input id="name" label="Default" variant="default" />
       <div style={{ marginBottom: "10px" }} />
-      <Input id="name" label="Biography" variant="underlined" />
+      <Input id="name" label="Underlined" variant="underlined" />
     </Form>
   );
 }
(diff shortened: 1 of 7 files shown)


packages/riipen-ui-docs/src/pages/components/input/input.md
```diff
@@ -35,6 +35,12 @@ The `labelWeight` prop can change the weight of the label.
 
 {{"demo": "pages/components/input/Weight.js"}}
 
+## Size
+
+The `size` prop changes the size of the input field.
```
ksmort, line 40: This is possibly the most Captain Obvious description ever 😅

packages/riipen-ui/src/components/Input.jsx
```diff
@@ -133,6 +134,18 @@ const Input = ({
...
+          padding: ${theme.spacing(1)}px;
+        }
+
+        .medium {
+          padding: ${theme.spacing(2)}px;
+        }
+
+        .large {
+          padding: ${theme.spacing(4)}px;
+        }
```
stredger, line 148: hmmm i wonder if we want to scale the text size instead of adding padding? The places i think we would use it would be something like the course wizard name input?

ksmort: Yeah I was wondering that too.  Any suggestions on font size? I'd assume `small` would be something like body1, but not sure how big (and bold?) large should get.
I'll give it a try and report back with screenshots!

packages/riipen-ui-docs/src/pages/components/input/Variant.jsx
```diff
@@ -6,9 +6,9 @@ import Input from "riipen-ui/components/Input";
 export default function Variant() {
   return (
     <Form>
-      <Input id="name" label="Biography" variant="default" />
+      <Input id="name" label="Default" variant="default" />
```
ksmort, line 9: Make these labels better

stredger (approved):

PR #188, opened 2021-04-07
Add icon prop to Input

ksmort:
## Description
Add icon prop to Input component. This renders an icon on the left side of the input.  Also update icon sizes based on `size` prop.
Update docs and tests.

## Notes
Open to adding support for flexibility of icons on either right or left side.
Based off https://github.com/riipen/ui/pull/187.

## Screenshots
Screenshot_2021-04-06 Input Riipen UI(3)

## Where to Start

### DIFF
--- a/packages/riipen-ui/src/components/Input.jsx
+++ b/packages/riipen-ui/src/components/Input.jsx
@@ -48,7 +48,7 @@ const Input = ({
   const componentClassName = clsx(
     error ? "error" : null,
     disabled ? "disabled" : null,
-    Icon ? "iconInput" : null,
+    Icon ? "iconPadding" : null,
     warning ? "warning" : null,
     focusVisible ? "focusVisible" : null,
     variant,
@@ -153,15 +153,15 @@ const Input = ({
           padding: 17px;
         }
 
-        .iconInput.small {
+        .iconPadding.small {
           padding-left: ${theme.spacing(9)}px;
         }
 
-        .iconInput.medium {
+        .iconPadding.medium {
           padding-left: ${theme.spacing(9)}px;
         }
 
-        .iconInput.large {
+        .iconPadding.large {
           padding-left: ${theme.spacing(12)}px;
         }
 
--- a/packages/riipen-ui/src/components/Input.jsx
+++ b/packages/riipen-ui/src/components/Input.jsx
@@ -13,6 +13,7 @@ const Input = ({
   disabled,
   error,
   hint,
+  icon: Icon,
   label,
   labelColor,
   labelWeight,
@@ -47,6 +48,7 @@ const Input = ({
   const componentClassName = clsx(
     error ? "error" : null,
     disabled ? "disabled" : null,
+    Icon ? "iconInput" : null,
     warning ? "warning" : null,
     focusVisible ? "focusVisible" : null,
     variant,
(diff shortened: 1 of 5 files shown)


packages/riipen-ui/src/components/Input.jsx
```diff
@@ -146,6 +175,16 @@ const Input = ({
...
+
+        .underlined {
+          background-color: transparent;
+          border: none;
+          border-bottom: 1px solid rgba(0, 0, 0, 0.23);
+        }
+
+        .warning {
+          border-color: ${theme.palette.warning.main};
+        }
```
ksmort, line 187: Organizing

packages/riipen-ui/src/components/Input.jsx
```diff
@@ -75,6 +77,11 @@ const Input = ({
         aria-disabled={disabled}
         {...other}
       />
+      {Icon && (
+        <span className={clsx("icon", size)}>
+          {typeof Icon === "function" ? <Icon /> : Icon}
```
ksmort, line 82: Ripped this from Button

packages/riipen-ui/src/components/Input.jsx
```diff
@@ -47,6 +48,7 @@ const Input = ({
   const componentClassName = clsx(
     error ? "error" : null,
     disabled ? "disabled" : null,
+    Icon ? "iconInput" : null,
```
ksmort, line 51: Not stoked on this class name but couldn't think of a better idea to get that left padding

packages/riipen-ui/src/components/Input.jsx
```diff
@@ -125,14 +132,36 @@ const Input = ({
...
+          color: ${theme.palette.grey[500]};
+          left: ${theme.spacing(4)}px;
+          position: absolute;
         }
 
-        .warning {
-          border-color: ${theme.palette.warning.main};
+        .icon.small {
+          font-size: ${theme.typography.body2.fontSize};
+          padding: 7px;
```
ksmort, line 143: Also not stoked on the custom padding to get it to align nicely left :/

packages/riipen-ui/src/components/Input.jsx
jordanell: Maybe just `iconPadding`? :shrug:

packages/riipen-ui/src/components/Input.jsx
```diff
@@ -125,14 +132,36 @@ const Input = ({
...
         }
 
-        .underlined {
-          background-color: transparent;
-          border: none;
-          border-bottom: 1px solid rgba(0, 0, 0, 0.23);
+        .icon {
+          color: ${theme.palette.grey[500]};
+          left: ${theme.spacing(4)}px;
+          position: absolute;
```
jordanell, line 138: I guess the question is should we allow icons to be any color people want? I think this is fine though. Unless anyone else wants to change this.

packages/riipen-ui/src/components/Input.jsx
jordanell: Meh. I think it's fine.

This CSS is assuming the icon size is settable using CSS `font-size`. This might not be true for all iconography libraries. But I am not sure there is a better solution. If people ever use an icon outside of normal svgs they can just put in their own CSS to make the sizing work.

jordanell (approved):

packages/riipen-ui/src/components/Input.jsx
ksmort: Yeah, wasn't sure on forcing the colour

PR #189, opened 2021-04-07
Add meta prop to Input component

ksmort:
## Description
Add `meta` prop to Input. This displays things on the bottom right of the Input field.
Update some logic with rendering error/warning, so that it works nicely with meta.
Update tests/docs.

## Screenshots
Screenshot_2021-04-07 Input Riipen UI(4)

## Where to Start

### DIFF
--- a/packages/riipen-ui/src/components/Input.jsx
+++ b/packages/riipen-ui/src/components/Input.jsx
@@ -17,6 +17,7 @@ const Input = ({
   label,
   labelColor,
   labelWeight,
+  meta,
   multiline,
   required,
   size,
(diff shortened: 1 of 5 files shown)


packages/riipen-ui/src/components/Input.jsx
```diff
@@ -170,6 +187,10 @@ const Input = ({
           padding-top: ${theme.spacing(1)}px;
         }
 
+        .medium {
```
ksmort, line 190: Need this for icon in input (accidentally removed)!

jordanell (approved):
Really nice!

PR #190, opened 2021-04-07
Add end prop to Input component, rename icon to start

ksmort:
## Description
Update `icon` prop to be `start`.  Add `end` prop, which displays content on the right side of the input, similar to the `start` prop.  Remove default icon colour, reorganize css.
Update docs and tests.

## Screenshots
Screenshot_2021-04-07 Input Riipen UI(5)

## Where to Start


packages/riipen-ui-docs/src/pages/components/input/input.md
```diff
@@ -29,11 +29,11 @@ The `multiline` prop transforms the input into a `textarea`.
 
 {{"demo": "pages/components/input/Multiline.js"}}
 
-## Icon
+## Start and End
```
ksmort, line 32: Not sure if I should call this "Controls"?

packages/riipen-ui/src/components/Input.jsx
```diff
@@ -204,6 +223,10 @@ const Input = ({
         .warning {
           border-color: ${theme.palette.warning.main};
         }
+
+        .wrapper {
+          position: relative;
```
ksmort, line 228: Add position relative so that the start/end position absolute is relative to the wrapper.

packages/riipen-ui/src/components/Input.jsx
```diff
@@ -11,16 +11,17 @@ import Typography from "./Typography";
 const Input = ({
   classes,
   disabled,
+  end,
```
jordanell, line 14: I wonder if we can be slightly more descriptive with these? Maybe just straight copy material-ui with endAdornment and startAdornment?

jordanell (changes requested):
Looks good. Just a suggestion around naming.

jordanell (approved):

PR #191, opened 2021-04-08
Checkbox chip

stredger:

### DIFF
--- a/packages/riipen-ui-docs/src/pages/components/checkbox/Variant.jsx
+++ b/packages/riipen-ui-docs/src/pages/components/checkbox/Variant.jsx
@@ -1,0 +1,38 @@
+import React from "react";
+
+import Form from "riipen-ui/components/Form";
+import Checkbox from "riipen-ui/components/Checkbox";
+
+export default function Usage() {
+  const style = {
+    marginBottom: "10px"
+  };
+
+  const [defaultState, setDefaultState] = React.useState(true);
+  const [chipState, setChipState] = React.useState(true);
+
+  return (
+    <Form>
+      {"asda"}
+      {defaultState}
+      {"asda"}
+      <Checkbox
+        checked={defaultState}
+        color="secondary"
+        onChange={e => setDefaultState(e.target.checked)}
+        id="defatul"
+        label="default"
+        variant="default"
+      />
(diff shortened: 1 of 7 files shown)


packages/riipen-ui-docs/src/pages/components/checkbox/Variant.jsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+
+import Form from "riipen-ui/components/Form";
+import Checkbox from "riipen-ui/components/Checkbox";
+
+export default function Usage() {
```
ksmort, line 6:
```suggestion
export default function Variant() {
```

packages/riipen-ui/src/components/Checkbox.jsx
```diff
@@ -80,23 +109,24 @@ const Checkbox = props => {
           min-height: 20px;
         }
 
+        .shift-label {
+          padding-left: 35px;
```
ksmort, line 113: Could use theme spacing

ksmort (commented):
Part of me feels it should be in Chip like you said, I guess with thinking how many variants for Checkbox will end up having.  Otherwise if it is going to live in Checkbox things look pretty 👌

jordanell:
@stredger I'm actually not a huge fan of this PR being in UI.

The UI kit has a checkbox component, and it has a chip component. I don't think we should conflate them in the base UI kit. If any consuming repo wants to use a chip as a checkbox, that should be handled inside of that repo.

stredger:
> @stredger I'm actually not a huge fan of this PR being in UI.
> 
> The UI kit has a checkbox component, and it has a chip component. I don't think we should conflate them in the base UI kit. If any consuming repo wants to use a chip as a checkbox, that should be handled inside of that repo.

Yea fair point. What i really want is a chip that is backed by an input so i can either:
1. Add a `component=input` or something to the Chip in the ui kit. (kinda has the same problem with this PR)
2. Do something like `<Chip><Checkbox classes={[ .hideCheckbox ]} /></Chip>` in web
3. Do something like `<Chip><input type="checkbox" className={.hideInput}></Chip>` in web and not use the ui kit at all for that type of input

PR #195, opened 2021-04-20
Update label prop type and remove labelWeight/labelColor

ksmort:
## Description
Update label prop type to be node so that it is not limited to strings.
Remove labelColor and labelWeight.  These can be passed into label prop in a component (ex. Typography).
Update docs/tests.

## Screenshots
 
Screenshot_2021-04-19 Input Riipen UI

## Where to Start


packages/riipen-ui/src/components/Input.jsx
```diff
@@ -62,11 +60,9 @@ const Input = ({
...
       {(label || hint) && (
         <InputLabel
-          color={labelColor}
           hint={hint}
           htmlFor={other.id || other.name}
           required={required}
-          weight={labelWeight}
         >
           {label}
         </InputLabel>
```
ksmort, line 68: One thing I'm a bit iffy on here is passing a custom InputLabel through via props.  This current code would look like `<InputLabel><InputLabel>{ .....} </InputLabel></InputLabel> `😬

jordanell: I think for right now, we leave `label` as a `string` only, not node. That would remove this concern. I think to truely switch this to a string or node, we need a slightly bigger PR which involves moving `<Typography>` out of the `InputLabel` component. Just not a change we need to do right now.

jordanell (changes requested):

packages/riipen-ui/src/components/Input.jsx
ksmort: Okay sounds good.  I'll close this PR!

PR #196, opened 2021-04-27
Chip Styles

jordanell:
## Description

1. Updated chip grey colors available
2. Updated chip disabled styles
3. Updated chip focus styles
4. Updated chip clickable styles
5. Updated tests / docs for chip

## Where to Start

The chip component. It's pretty heavily refactored to might be best to just examine the whole file rather than the diff.

### DIFF
--- a/packages/riipen-ui/src/components/Chip.jsx
+++ b/packages/riipen-ui/src/components/Chip.jsx
@@ -1,330 +1,399 @@
 import clsx from "clsx";
 import PropTypes from "prop-types";
-import React from "react";
+import React, { useState } from "react";
 import _JSXStyle from "styled-jsx/style";
 
-import ThemeContext from "../styles/ThemeContext";
+import useIsFocusVisible from "../utils/useIsFocusVisible";
 import withClasses from "../utils/withClasses";
+import withThemeContext from "../utils/withThemeContext";
 
-class Chip extends React.Component {
-  static displayName = "Chip";
+const Chip = ({
+  children,
+  classes,
+  color,
+  component: Component,
+  disabled,
+  hover,
+  icon,
+  label,
+  onClick,
+  size,
+  theme,
+  variant
+}) => {
+  const [focusVisible, setFocusVisible] = useState(false);
+  const { ref, isFocusVisible, onBlurVisible } = useIsFocusVisible();
 
-  static propTypes = {
-    /**
-     * The elements to render inside the pill.
-     */
-    children: PropTypes.node,
+  const clickable = !!onClick;
 
-    /**
-     * List of additional classes to apply to this component.
-     */
-    classes: PropTypes.array,
+  const className = clsx(
+    "root",
+    color,
+    disabled ? "disabled" : null,
+    size,
+    variant,
+    classes,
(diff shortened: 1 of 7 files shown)


packages/riipen-ui/src/components/Chip.jsx
```diff
@@ -1,330 +1,399 @@
...
+    onBlurVisible();
+  };
 
-    /**
-     * If `true`, the chip should be displayed in a disabled state.
-     */
-    disabled: PropTypes.bool,
+  const handleClick = () => {
+    if (onClick) onClick();
+  };
```
ksmort, line 48: Can onClick just get passed into the component?

packages/riipen-ui/src/components/Chip.jsx
```diff
@@ -1,330 +1,399 @@
...
+          background-color: ${theme.palette.grey[300]};
+          border-radius: 50px;
+          border: 1px solid;
+          border-color: transparent;
+          box-sizing: border-box;
+          color: ${theme.palette.text.secondary};
+          cursor: default;
+          display: inline-flex;
+          font-family: ${theme.typography.fontFamily};
+          font-size: 13px;
```
ksmort, line 78: Is it possible to use one of our preset font sizes here?

packages/riipen-ui/src/components/Chip.jsx
```diff
@@ -1,330 +1,399 @@
...
+          white-space: nowrap;
+        }
+        .root.xsmall {
+          padding: 1px 5px;
+        }
+        .root.small {
+          padding: 5px 10px;
+        }
+        .root.medium {
+          padding: 10px 15px;
```
ksmort, line 91: Can use theme.spacing for some of these

packages/riipen-ui/src/components/Chip.jsx
```diff
@@ -1,330 +1,399 @@
...
+        .root.clickable:hover,
+        .root.focusVisible,
+        .root.hover:hover {
+          background-color: transparent;
+          border: 1px solid ${theme.palette.text.primary};
+        }
+        .icon {
+          align-items: center;
+          display: flex;
+          padding-right: 10px;
```
ksmort, line 109: ditto theme spacing

packages/riipen-ui/src/components/Chip.jsx
```diff
@@ -1,330 +1,399 @@
...
+          align-items: center;
+          display: flex;
+          white-space: nowrap;
+        }
 
-    /**
-     * Action to perform when clicked.
-     */
-    onClick: PropTypes.func,
+        .primary {
```
ksmort, line 119: What's our rule on whitespace between css in the ui kit?  Some of these have newlines between them, some don't

ksmort (changes requested):

packages/riipen-ui/src/components/Chip.jsx
jordanell: We actually don't really have a rule. I just try to group them to what makes sense given the context.

ksmort (approved):
🙌

PR #197, opened 2021-04-27
Radio button cleanup

stredger:

### DIFF
--- a/packages/riipen-ui/src/components/RadioButton.jsx
+++ b/packages/riipen-ui/src/components/RadioButton.jsx
@@ -39,10 +39,7 @@ const RadioButton = ({
         size
       )}
     >
-      <label
-        className={clsx(checked && "checked", disabled && "disabled")}
-        htmlFor={other.id || other.name}
-      >
+      <label className={clsx(checked && "checked", disabled && "disabled")}>
         <input checked={checked} disabled={disabled} type="radio" {...other} />
         <Typography
           component="span"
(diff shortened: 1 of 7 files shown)


packages/riipen-ui/src/components/RadioButton.jsx
```diff
@@ -39,10 +39,7 @@ const RadioButton = ({
         size
       )}
     >
-      <label
-        className={clsx(checked && "checked", disabled && "disabled")}
-        htmlFor={other.id || other.name}
```
ksmort, line 44: Do we not need htmlFor?

packages/riipen-ui/src/components/RadioButtonGroup.jsx
```diff
@@ -61,7 +61,7 @@ const RadioButtonGroup = ({
   };
 
   const childrenWithProps = React.Children.map(
-    children,
+    children || [],
```
ksmort, line 64: Looks like children is a required prop, so if we want to do this (also maybe as a default prop?), should update the prop type

packages/riipen-ui/src/components/RadioGroup.jsx
```diff
@@ -7,104 +7,118 @@ import withClasses from "../utils/withClasses";
...
-    /**
-     * A warning to display below the input.
-     */
-    warning: PropTypes.node
-  };
-
-  handleChange = e => {
-    const { onChange } = this.props;
-
+const RadioGroup = ({
```
ksmort, line 10: 🙌

ksmort (changes requested):
Overall lgtm!  Couple nitpicks

packages/riipen-ui/src/components/RadioButton.jsx
stredger: this one wrapps the input so the label takes the child input, which is always what we want as its the text on the button

ksmort (approved):

PR #198, opened 2021-04-27
Fix name prop in radio groups

stredger:
## Description

Fix passing busted name prop in radio groups


packages/riipen-ui/src/components/RadioButtonGroup.jsx
```diff
@@ -66,7 +65,6 @@ const RadioButtonGroup = ({
       child &&
       React.cloneElement(child, {
         checked: value === child.props.value,
-        name: `${name}-${child.props.value}`,
```
stredger, line 69: this should not have been a thing, my b

jordanell (approved):

PR #206, opened 2021-05-06
Update tooltip top/bottom padding to be a bit smaller in small size

ksmort:
## Description
Update small tooltip top/bottom padding to be a bit smaller.

## Screenshots
Screen Shot 2021-05-06 at 12 48 18 PM


stredger (approved):
so crisp

PR #214, opened 2021-06-01
[PLT-7534] Add color prop to Avatar

ericauv:
## Description
* Add a `color` prop to `<Avatar>`, defaults to "grey" and can be one of "grey" or "white"
## Notes
* required for https://github.com/riipen/platform-web/pull/4242
## Screenshots
Avatar with background white, others with default grey background
Screen Shot 2021-06-01 at 3 08 07 PM

### DIFF
--- a/packages/riipen-ui/src/components/Avatar.jsx
+++ b/packages/riipen-ui/src/components/Avatar.jsx
@@ -28,7 +28,7 @@ class Avatar extends React.Component {
     /**
      * The color of the background for the Avatar
      */
-    backgroundColor: PropTypes.oneOf(["grey", "white"]),
+    backgroundColor: PropTypes.oneOf(["grey200", "white"]),
 
     /**
      * The height and width size to render the Avatar at.
@@ -48,7 +48,7 @@ class Avatar extends React.Component {
 
   static defaultProps = {
     classes: [],
-    backgroundColor: "grey",
+    backgroundColor: "grey200",
     size: "96px",
     variant: "circle"
   };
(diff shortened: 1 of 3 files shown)


jordanell (commented):
@ericauv I'm not entirely convinced this is the way to go. We already have `classes` which can be set to something with a specific background color. Just curious why we want to make it an explicit prop with this PR.

ericauv:
> @ericauv I'm not entirely convinced this is the way to go. We already have `classes` which can be set to something with a specific background color. Just curious why we want to make it an explicit prop with this PR.

Most of our other `ui` components have a color prop, I think this change makes sense because it makes the use of our ui components more consistent.
I think we should either always use the color prop to set background-color of our ui components, or always use classes --> I think using the color prop is an easier/cleaner dev experience

jordanell:
@ericauv that's fair. I think we should call this one `backgroundColor` though to distinguish it from other foreground color props.

packages/riipen-ui/src/components/Avatar.jsx
```diff
@@ -25,6 +25,11 @@ class Avatar extends React.Component {
      */
     classes: PropTypes.array,
 
+    /**
+     * The color of the background for the Avatar
+     */
+    backgroundColor: PropTypes.oneOf(["grey", "white"]),
```
jordanell, line 31: Sorry I should have caught this earlier. When we use greys in our props, we have the standard of using the fully qualified grey like "grey200", "grey600", or "greyA400", so I think in this case, we should probably use `grey200`.

jordanell (changes requested):

jordanell (approved):

PR #216, opened 2021-06-04
Mobile menu for UI Riipen page

ksmort:
## Description
Add Menu for "mobile" view, as the left-hand menu disappears at smaller screen sizes.
Mostly a  proof of concept PR, definitely needs more work and cleaning up.

## Notes
Mostly a  proof of concept PR, definitely needs more work and cleaning up.

## Screenshots
Screenshot_2021-06-03 App Bar API Riipen UI
Screenshot 2021-06-04 at 16-50-09 Badge Riipen UI

## Where to Start
packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx

### DIFF
--- a/packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
+++ b/packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
@@ -25,6 +25,7 @@ const MarkdownPage = ({ req, path, reqSource, title }) => {
   const [menuOpen, setMenuOpen] = React.useState(false);
 
   const handleMenuOpen = (event) => {
+    document.body.style.overflow = menuOpen ? "auto" : "hidden";
     setAnchorState(menuOpen ? null : event.currentTarget);
     setMenuOpen(!menuOpen);
   };
@@ -55,29 +56,28 @@ const MarkdownPage = ({ req, path, reqSource, title }) => {
       <Head>
         <title>{title ? `${title} | ` : ""}Riipen UI</title>
       </Head>
-      <AppBar>
+      <AppBar classes={["appBar"]}>
         <Link href="/" color="inherit">
           Riipen-UI
         </Link>
         <div className="menuMobile">
-          <React.Fragment>
-            <ButtonIcon color="white" onClick={handleMenuOpen}>
-              <FontAwesomeIcon icon={menuOpen ? faTimes : faBars} />
-            </ButtonIcon>
-            <Popover
-              anchorEl={anchorState}
-              anchorPosition={{
-                vertical: "bottom",
-                horizontal: "center"
-              }}
-              closeOnScrolled={false}
-              fullWidth
-              isOpen={menuOpen}
-              lockScroll
-            >
-              <Menu />
-            </Popover>
-          </React.Fragment>
+          <ButtonIcon color="white" onClick={handleMenuOpen}>
+            <FontAwesomeIcon icon={menuOpen ? faTimes : faBars} />
+          </ButtonIcon>
+          <Popover
+            anchorEl={anchorState}
+            anchorPosition={{ horizontal: "center", vertical: "bottom" }}
+            classes={["menuPopover"]}
+            contentPosition={{
+              horizontal: "center",
+              vertical: "top"
+            }}
+            closeOnScrolled={false}
+            fullWidth
+            isOpen={menuOpen}
+          >
+            <Menu />
+          </Popover>
         </div>
       </AppBar>
       <div className="page">
@@ -122,6 +122,9 @@ const MarkdownPage = ({ req, path, reqSource, title }) => {
         </div>
       </div>
       <style jsx>{`
+        :global(.appBar) {
+          justify-content: space-between;
+        }
         .page {
           display: flex;
           justify-content: space-between;
@@ -144,7 +147,11 @@ const MarkdownPage = ({ req, path, reqSource, title }) => {
         .menuMobile {
           display: none;
         }
-        .menuMobile :global(.popover) {
+        :global(.menuPopover) {
+          margin: ${theme.spacing(3)}px ${theme.spacing(2)}px;
+          max-height: calc(100% - ${theme.spacing(13)}px) !important;
+          max-width: calc(100% - ${theme.spacing(4)}px) !important;
+          padding: ${theme.spacing(2)}px;
           overflow-y: auto;
         }
         .toc {
@@ -161,7 +168,6 @@ const MarkdownPage = ({ req, path, reqSource, title }) => {
 
           .menuMobile {
             display: initial;
-            margin-left: auto;
           }
         }
         .footer {
--- a/packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
+++ b/packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
@@ -1,6 +1,13 @@
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faBars } from "@fortawesome/free-solid-svg-icons";
 import Head from "next/head";
 import PropTypes from "prop-types";
 import React from "react";
+import AppBar from "riipen-ui/components/AppBar";
+import ButtonIcon from "riipen-ui/components/ButtonIcon";
+import Divider from "riipen-ui/components/Divider";
+import Popover from "riipen-ui/components/Popover";
+import ThemeContext from "riipen-ui/styles/ThemeContext";
 
 import constants from "src/constants";
 import Demo from "src/modules/components/Demo";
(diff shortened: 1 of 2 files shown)


packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
```diff
@@ -11,141 +18,164 @@ import Menu from "src/modules/components/Menu";
 import TOC from "src/modules/components/TOC";
 import { getSections } from "src/utils/parseMarkdown";
 
-import AppBar from "riipen-ui/components/AppBar";
-import Divider from "riipen-ui/components/Divider";
-import ThemeContext from "riipen-ui/styles/ThemeContext";
+const MarkdownPage = ({ req, path, reqSource, title }) => {
+  const theme = React.useContext(ThemeContext);
```
ksmort, line 22: Convert to function component for the useState stuff

packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
```diff
@@ -11,141 +18,164 @@ import Menu from "src/modules/components/Menu";
...
+              <FontAwesomeIcon icon={faBars} />
+            </ButtonIcon>
+            <Popover
+              anchorEl={anchorState}
+              closeOnScrolled={false}
+              fullWidth
+              isOpen={menuOpen}
+              lockScroll
+            >
+              <Menu />
```
ksmort, line 74: Lazily just threw current Menu into a Popover... Could use Menu/MenuItem potentially 😅

packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
```diff
@@ -11,141 +18,164 @@ import Menu from "src/modules/components/Menu";
...
+          width: 100%;
+        }
+        .menu {
+          padding: ${theme.spacing(2)}px;
+        }
+        .menuMobile {
+          display: none;
+        }
+        .menuMobile :global(.popover) {
+          overflow-y: auto;
```
ksmort, line 144: A bit of hackery to make the popover scrolle :/

packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
```diff
@@ -11,141 +18,164 @@ import Menu from "src/modules/components/Menu";
...
           }
-          .footer {
-            margin: ${theme.spacing(10)}px 0;
-          }
-          .footer :global(footer) {
-            margin-top: ${theme.spacing(10)}px;
+
+          .menuMobile {
+            display: initial;
+            margin-left: auto;
```
ksmort, line 160: More hackery to make the menu bars icon display on the far right of the app bar (just passing in classes prop of AppBar wasn't working

packages/riipen-ui-docs/src/modules/components/Menu.jsx
```diff
@@ -54,7 +55,9 @@ const Menu = () => {
           <li key={i}>
             <Accordion {...parent.props}>
               <AccordionSummary icon={Icon} iconProps={{ size: "small" }}>
-                {parent.name}
+                <Typography color="initial" fontWeight="bold" variant="h5">
```
ksmort, line 58: Make sure the parent.name doesn't randomly become white when it goes to mobile?  Some `primary` colour overrides the `inherit` colour for some reason.

packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
jordanell: Nice! I tried this out in the UI, but there seems to be no way to actually close the menu.

Screenshot from 2021-06-04 08-18-53

packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
jordanell: Can we pass `classes` to the popover? If not maybe we should add that ability instead of this.

packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
jordanell: We should fix `classes` to app bar if they aren't woring instead of this. I imagine you need to pass a CSS class to Appbar that is something like:

```css
.appbar {
  display: flex;
  justify-content: space-between;
}
```

jordanell (changes requested):
Nice rogue day! Some suggestions to clean this up a bit.

Some other suggestions:

- Add some padding between the popover and the sides of the screen, maybe 5px?
- Add some padding between the popover and the top of the screen so it looks like it is appearing below the appbar

packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
ksmort: Figured it out, need :global(.classToPassToAppBar) in MarkdownPage!

packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
```diff
@@ -11,141 +18,174 @@ import Menu from "src/modules/components/Menu";
...
+  const [menuOpen, setMenuOpen] = React.useState(false);
 
-class MarkdownPage extends React.Component {
-  static propTypes = {
-    path: PropTypes.string.isRequired,
-    req: PropTypes.func.isRequired,
-    reqSource: PropTypes.func,
-    title: PropTypes.string
+  const handleMenuOpen = (event) => {
+    document.body.style.overflow = menuOpen ? "auto" : "hidden";
```
ksmort, line 28: Force lock scroll.
I think popover might be supposed to do this but it looks like it only works on mount. @bolinkd can you confirm if that's desired behaviour to only lock on mount (and not when anchorEl/lockScroll props change)? You are the Popover guy in my mind :P 
I think Popover can use a Convert to Function facelift anyways :)

packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
```diff
@@ -11,141 +18,174 @@ import Menu from "src/modules/components/Menu";
...
+          <ButtonIcon color="white" onClick={handleMenuOpen}>
+            <FontAwesomeIcon icon={menuOpen ? faTimes : faBars} />
+          </ButtonIcon>
+          <Popover
+            anchorEl={anchorState}
+            anchorPosition={{ horizontal: "center", vertical: "bottom" }}
+            classes={["menuPopover"]}
+            contentPosition={{
+              horizontal: "center",
+              vertical: "top"
```
ksmort, line 73: New props to make it be in better position

packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
```diff
@@ -11,141 +18,174 @@ import Menu from "src/modules/components/Menu";
...
+      <Head>
+        <title>{title ? `${title} | ` : ""}Riipen UI</title>
+      </Head>
+      <AppBar classes={["appBar"]}>
+        <Link href="/" color="inherit">
+          Riipen-UI
+        </Link>
+        <div className="menuMobile">
+          <ButtonIcon color="white" onClick={handleMenuOpen}>
+            <FontAwesomeIcon icon={menuOpen ? faTimes : faBars} />
```
ksmort, line 65: Differentiate between open/closed. This might be overkill. Lmk!

packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
```diff
@@ -11,141 +18,174 @@ import Menu from "src/modules/components/Menu";
...
+        .menu {
+          padding: ${theme.spacing(2)}px;
+        }
+        .menuMobile {
+          display: none;
+        }
+        :global(.menuPopover) {
+          margin: ${theme.spacing(3)}px ${theme.spacing(2)}px;
+          max-height: calc(100% - ${theme.spacing(13)}px) !important;
+          max-width: calc(100% - ${theme.spacing(4)}px) !important;
```
ksmort, line 153: Some overriding of popover.  Should I use linkedStyles (css.resolve) instead? I was lazy and just used importants

packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
```diff
@@ -11,141 +18,174 @@ import Menu from "src/modules/components/Menu";
...
+        }
+        .page {
+          display: flex;
+          justify-content: space-between;
+          padding-top: 22px;
+          position: relative;
+          top: 50px;
+          width: 100%;
+        }
+        .page :global(*:target) {
```
jordanell, line 136: I think instead of all these `:global` tags, we actually want to use https://github.com/vercel/styled-jsx#the-resolve-tag.

There's a bunch of examples of that being used in the actually `riipen-ui` package for reference.

jordanell (changes requested):
Looks good, but I think we should get rid of the `:global` stuff. See my comment.

packages/riipen-ui-docs/src/modules/components/MarkdownPage.jsx
bolinkd: I believe the original intention is that the scroll of the window behind is disabled so that the scrolling the user does is captured by the popover (it might be the case where we want that interaction configurable)

jordanell (approved):

PR #217, opened 2021-06-04
Convert Tooltip to Function Component and Add Unit Tests

ksmort:
## Description
Update Tooltip so that it is a function component.
Add tests + snapshots.
Updating to a function component seems to have fixed the Samsung focus bug.

## Where to Start
components/Tooltip

### DIFF
--- a/packages/riipen-ui/src/components/Tooltip.jsx
+++ b/packages/riipen-ui/src/components/Tooltip.jsx
@@ -1,5 +1,5 @@
 import PropTypes from "prop-types";
-import React, { useRef, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import css from "styled-jsx/css";
 
 import ThemeContext from "../styles/ThemeContext";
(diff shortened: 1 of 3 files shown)


packages/riipen-ui/src/components/Tooltip.jsx
```diff
@@ -8,130 +8,35 @@ import withClasses from "../utils/withClasses";
...
+}) => {
+  const theme = React.useContext(ThemeContext);
+  const [isOpen, setIsOpen] = useState(isControlledByProps ? open : false);
+  const tooltipRootRef = useRef(null);
+
+  useEffect(() => {
+    if (isControlledByProps) {
+      setIsOpen(open);
+    }
+  }, [isControlledByProps, open]);
```
ksmort, line 37: Just update state to open prop as needed. Cleans up the logic down the line a bit.

packages/riipen-ui/src/components/Tooltip.jsx
```diff
@@ -406,97 +311,66 @@ class Tooltip extends React.Component {
...
-  getOppositeDirection = (direction1, direction2, value) => {
-    if (value === direction1) {
-      return direction2;
-    } else if (value === direction2) {
-      return direction1;
+  const getOppositeDirection = (direction, oppositeDirection, value) => {
+    if (value === direction) {
+      return oppositeDirection;
+    } else if (value === oppositeDirection) {
+      return direction;
```
ksmort, line 318: Better names. Tbh I still have no idea what past-me was doing here

packages/riipen-ui/src/components/Tooltip.jsx
```diff
@@ -406,97 +311,66 @@ class Tooltip extends React.Component {
...
     }
     return value;
   };
 
-  blur = () => {
+  const blur = () => {
```
ksmort, line 323: I think if there is a focus error again, this is the perpetrator. But it seems okay as a function component?

jordanell (approved):
All looks good to me!

stredger (approved):

PR #219, opened 2021-06-05
Convert Popover to function component

ksmort:
## Description
Convert Popover to be a function component.
Add basic test file.

## Notes
TODO: Beef up tests.

## Where to Start
components/Popover

### DIFF
--- a/packages/riipen-ui/src/components/Popover.jsx
+++ b/packages/riipen-ui/src/components/Popover.jsx
@@ -153,13 +153,18 @@ const Popover = ({
         return;
       }
       handleClose();
-    } else if (event.type === "keydown" && event.key === "Escape") {
+    } else if (
+      ["resize", "scroll"].includes(event.type) ||
+      (event.type === "keydown" && event.key === "Escape")
+    ) {
       handleClose();
     } else if (event.type === "keydown" && onKeyDown) {
       // stop this event from triggering other keydown listeners
       event.stopImmediatePropagation();
 
       onKeyDown(event);
+    } else {
+      handleClose();
     }
   };
 
@@ -291,7 +296,7 @@ Popover.propTypes = {
   /**
    * The content of the component
    */
-  children: PropTypes.object,
+  children: PropTypes.node,
 
   /**
    * Array of additional CSS classes to use.
--- a/packages/riipen-ui/src/components/Popover.jsx
+++ b/packages/riipen-ui/src/components/Popover.jsx
@@ -15,14 +15,14 @@ import {
 } from "../utils";
 import withClasses from "../utils/withClasses";
 
-const isDecendant = (parent, child) => {
+const isDescendant = (parent, child) => {
   const node = child.parentNode;
   if (node == null) {
     return false;
   } else if (node === parent) {
     return true;
   }
-  return isDecendant(parent, node);
+  return isDescendant(parent, node);
 };
 
 const Popover = ({
@@ -147,8 +147,8 @@ const Popover = ({
       if (
         contentRef.current === target ||
         anchorEl === target ||
-        isDecendant(contentRef.current, target) ||
-        isDecendant(anchorEl, target)
+        isDescendant(contentRef.current, target) ||
+        isDescendant(anchorEl, target)
       ) {
         return;
       }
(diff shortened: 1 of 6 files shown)


ericauv (approved):
It _looks_ good -- We'll definitely want to do some heavy qa + should involve @Kyubinhan in qa too since he's recently done a deep dive into this component

ksmort:
> It _looks_ good -- We'll definitely want to do some heavy qa + should involve @Kyubinhan in qa too since he's recently done a deep dive into this component

Adding @Kyubinhan as reviewer!

packages/riipen-ui/src/components/Popover.jsx
```diff
@@ -274,140 +120,140 @@ class Popover extends React.Component {
...
-            {children}
-          </Component>
-        )}
-        <style jsx>{`
+  useEffect(() => {
+    if (isOpen) {
+      // need to use requestAnimationFrame in order for popover to get proper position on first open
+      window.requestAnimationFrame(setPositioningStyle);
+    }
+  }, [anchorEl, anchorPosition, contentPosition, isOpen]);
```
Kyubinhan, line 243: I am not sure if this matches with the previous logic. `isOpen` needs not to be checked whether changes occurred (just want to if that is true). Looks like this logic is also missing `const anchorChange =
      prevProps.anchorEl === null && this.props.anchorEl !== null;`

packages/riipen-ui/src/components/Popover.jsx
```diff
@@ -15,206 +15,53 @@ import {
...
-    if (
-      anchorChange ||
-      openChange ||
-      contentPositionChange ||
-      anchorPositionChange
-    ) {
-      // need to use requestAnimationFrame in order for popover to get proper position on first open
-      window.requestAnimationFrame(this.setPositioningStyle);
-    }
+const isDecendant = (parent, child) => {
```
Kyubinhan, line 18: Little typo? `isDecendant` -> `isDescendant`

Kyubinhan (commented):
Can you take a look at the tooltip visibility in UI docs? when `keep showing tooltip on leave` was checked and the tooltip was opened I was able to close the tooltip by clicking on somewhere and but it doesn't anymore :(

packages/riipen-ui/src/components/Popover.jsx
ksmort: Y'know, I just want to make sure my constants are up to Gordon Ramsay standards :P

packages/riipen-ui/src/components/Popover.test.jsx
```diff
@@ -0,0 +1,274 @@
...
+      window.addEventListener = jest.fn((event, callback) => {
+        events[event] = callback;
+      });
+      props.onClose = jest.fn();
+      mount(<Popover {...props} />);
+
+      events.keydown({
+        target: props.anchorEl,
+        type: "mousedown"
+      });
```
ksmort, line 236: Bit of a hack to call the listeners.  Open to better methods!

packages/riipen-ui/src/components/Popover.jsx
ksmort: Should be fixed now?

packages/riipen-ui/src/components/Popover.jsx
```diff
@@ -274,140 +120,145 @@ class Popover extends React.Component {
...
+      window.removeEventListener("keydown", handleCloseEvent);
 
-  resetParentScroll() {
-    this.state.scrollContainer.style.overflow = this.state.scrollContainerStyle;
-  }
+      manageScrollableParentEventListeners(true);
+    };
+  }, []);
+
+  useEffect(() => {
```
Kyubinhan, line 220: I think it would be nice if we leave comments for each useEffect. For example we can copy this comment from the previous Popover component.
```
// Find the outer most scrollable content and remove the scrollbar
```

Kyubinhan (commented):
Hmm were you able to test the `keep showing tooltip on leave` issue in Tooltip visibility? I still can't close the tooltip :(

ksmort:
@Kyubinhan @ericauv Giving up on converting to function component, but leaving the unit tests and small prop change to children of type object -> type node so that don't need to wrap tooltip in a fragment when passing it through.

PR #229, opened 2021-08-23
[PLT-8224] Add tooltip prop to RadioGroup

imogenem:
## Description
We are updating the Project Wizard to include descriptive Tooltips. Similar to how `FormGroup` was updated in `platform-web` to include the tooltip as a prop, we should update RadioGroup so it can pass a tooltip node.

Related PR: [Project Library Tooltips](https://github.com/riipen/platform-web/pull/4609)

## Screenshots
Added in text to show placement of where the tooltip would go

Screen Shot 2021-08-11 at 5 34 44 PM

### DIFF
--- a/packages/riipen-ui/src/components/RadioGroup.jsx
+++ b/packages/riipen-ui/src/components/RadioGroup.jsx
@@ -15,7 +15,7 @@ const RadioGroup = ({
   labelProps,
   onChange,
   required,
-  tooltip,
+  suffix,
   value,
   warning,
   ...other
@@ -40,13 +40,13 @@ const RadioGroup = ({
   return (
     <React.Fragment>
       <fieldset>
-        {(label || hint || tooltip) && (
+        {(label || hint || suffix) && (
           <InputLabel
             hint={hint}
             htmlFor={other.name}
             required={required}
             {...labelProps}
-            tooltip={tooltip}
+            suffix={suffix}
           >
             {label}
           </InputLabel>
@@ -114,9 +114,9 @@ RadioGroup.propTypes = {
   required: PropTypes.bool,
 
   /**
-   * Any tooltip info to display beside the label.
+   * Any suffix to display beside the label.
    */
-  tooltip: PropTypes.node,
+  suffix: PropTypes.node,
 
   /**
    * The value of the selected radio button.
--- a/packages/riipen-ui/src/components/RadioGroup.jsx
+++ b/packages/riipen-ui/src/components/RadioGroup.jsx
@@ -15,6 +15,7 @@ const RadioGroup = ({
   labelProps,
   onChange,
   required,
+  tooltip,
   value,
   warning,
   ...other
@@ -39,12 +40,13 @@ const RadioGroup = ({
   return (
     <React.Fragment>
       <fieldset>
-        {(label || hint) && (
+        {(label || hint || tooltip) && (
           <InputLabel
             hint={hint}
             htmlFor={other.name}
             required={required}
             {...labelProps}
+            tooltip={tooltip}
           >
             {label}
           </InputLabel>
@@ -94,7 +96,7 @@ RadioGroup.propTypes = {
   label: PropTypes.node,
 
   /**
-   * Any props to pass through to the InputLabel
+   * Any props to pass through to the InputLabel.
    */
   labelProps: PropTypes.object,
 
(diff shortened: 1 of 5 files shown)


bolinkd:
Hmm, im not sure how I feel about naming here. We dont "have" to provide a tooltip there, its can be any node we want to render after the text. Sadly, I can not currently think of a better name. We want to keep the names generic so that if we want to put something other than a tooltip after the text it would be fine.

I guess the other thought is does this have the be added to the UI Kit, we could just do it where we just inline two different DOM elements in the platform-web rather than passing one in as props.

@stredger any suggestions on better naming/if we should do this as part of the UI Kit?

stredger:
> Hmm, im not sure how I feel about naming here. We dont "have" to provide a tooltip there, its can be any node we want to render after the text. Sadly, I can not currently think of a better name. We want to keep the names generic so that if we want to put something other than a tooltip after the text it would be fine.
> 
> I guess the other thought is does this have the be added to the UI Kit, we could just do it where we just inline two different DOM elements in the platform-web rather than passing one in as props.
> 
> @stredger any suggestions on better naming/if we should do this as part of the UI Kit?

Yea I ran against this when i added something similar to the input. The problem we are trying to solve is we have 4 things we want to have on an input.
1. Label
2. Hint
3. Tooltip
4. Required star

The star is appended after the label, but we want the tooltip to be After both the label and the required star. This makes it difficult to just inline stuff into the label prop (we also want inputs to look the same). I agree that "tooltip" its kind of an awkward name as it can be any react content. Maybe "suffix" or something would be better? Anyways if someone can think of a better solution to what i described above I think that is the root problem

imogenem:
> > Hmm, im not sure how I feel about naming here. We dont "have" to provide a tooltip there, its can be any node we want to render after the text. Sadly, I can not currently think of a better name. We want to keep the names generic so that if we want to put something other than a tooltip after the text it would be fine.
> > I guess the other thought is does this have the be added to the UI Kit, we could just do it where we just inline two different DOM elements in the platform-web rather than passing one in as props.
> > @stredger any suggestions on better naming/if we should do this as part of the UI Kit?
> 
> Yea I ran against this when i added something similar to the input. The problem we are trying to solve is we have 4 things we want to have on an input.
> 
> 1. Label
> 2. Hint
> 3. Tooltip
> 4. Required star
> 
> The star is appended after the label, but we want the tooltip to be After both the label and the required star. This makes it difficult to just inline stuff into the label prop (we also want inputs to look the same). I agree that "tooltip" its kind of an awkward name as it can be any react content. Maybe "suffix" or something would be better? Anyways if someone can think of a better solution to what i described above I think that is the root problem

I gave the inline styling a bit of a try when I first started this ticket, but ended up thinking that adding the `suffix/tooltip` prop would be cleaner for `platform-web` - and it seems possible that `RadioGroup` might need this prop elsewhere in the future to add more tooltips.

bolinkd:
> > > Hmm, im not sure how I feel about naming here. We dont "have" to provide a tooltip there, its can be any node we want to render after the text. Sadly, I can not currently think of a better name. We want to keep the names generic so that if we want to put something other than a tooltip after the text it would be fine.
> > > I guess the other thought is does this have the be added to the UI Kit, we could just do it where we just inline two different DOM elements in the platform-web rather than passing one in as props.
> > > @stredger any suggestions on better naming/if we should do this as part of the UI Kit?
> > 
> > 
> > Yea I ran against this when i added something similar to the input. The problem we are trying to solve is we have 4 things we want to have on an input.
> > 
> > 1. Label
> > 2. Hint
> > 3. Tooltip
> > 4. Required star
> > 
> > The star is appended after the label, but we want the tooltip to be After both the label and the required star. This makes it difficult to just inline stuff into the label prop (we also want inputs to look the same). I agree that "tooltip" its kind of an awkward name as it can be any react content. Maybe "suffix" or something would be better? Anyways if someone can think of a better solution to what i described above I think that is the root problem
> 
> I gave the inline styling a bit of a try when I first started this ticket, but ended up thinking that adding the `suffix/tooltip` prop would be cleaner for `platform-web` - and it seems possible that `RadioGroup` might need this prop elsewhere in the future to add more tooltips.

I would be more in favour of renaming the prop to something more generic like suffix instead then

packages/riipen-ui/src/components/RadioGroup.jsx
```diff
@@ -39,12 +40,13 @@ const RadioGroup = ({
...
     <>
       <fieldset>
-        {(label || hint) && (
+        {(label || hint || suffix) && (
           <InputLabel
             hint={hint}
             htmlFor={other.name}
             required={required}
             {...labelProps}
+            suffix={suffix}
```
bolinkd, line 49: i believe we should swap order here, based on what im reading, i think label props should overwrite any existing props

bolinkd (approved):

PR #231, opened 2021-09-01
[PLT-8291]: Update tests to use invoke instead of simulate 

ksmort:
## Description
Update rest of ui kit tests to use invoke instead of simulate.

### DIFF
--- a/packages/riipen-ui/src/components/RadioButtonGroup.test.jsx
+++ b/packages/riipen-ui/src/components/RadioButtonGroup.test.jsx
@@ -140,7 +140,7 @@ describe("<RadioButtonGroup>", () => {
         .find("input")
         .invoke("onChange")({ target: { value: "a" } });
 
-      expect(onChange).toHaveBeenCalledTimes(1);
+      expect(onChange).toHaveBeenCalledWith({ target: { value: "a" } }, "a");
     });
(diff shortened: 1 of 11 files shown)


packages/riipen-ui/src/components/RadioButtonGroup.test.jsx
```diff
@@ -138,7 +138,7 @@ describe("<RadioButtonGroup>", () => {
       wrapper
         .find("RadioButton")
         .find("input")
-        .simulate("change");
+        .invoke("onChange")({ target: { value: "a" } });
```
ericauv, line 141: could do something similar to `assert.deepEqual(onChange.event, {target:{value:'a'})`

packages/riipen-ui/src/components/RadioGroup.test.jsx
```diff
@@ -124,7 +124,7 @@ describe("<RadioGroup>", () => {
       wrapper
         .find("Radio")
         .find("input")
-        .simulate("change");
+        .invoke("onChange")({ target: { value: "a" } });
```
ericauv, line 127: same here re: checking that onChange is called with this event

ericauv (approved):

PR #235, opened 2021-10-01
Make sure to pass on the event object to onFocus & onBlur props in Input

Kyubinhan:
## Description
Forgot to pass on the event object to these event handler props. :(


ksmort (commented):
🚀

ksmort (approved):

PR #236, opened 2021-10-29
Split table generator into 2 components to avoid whole table re-renders

stredger:
## Description

Since the hover state was stored at the table level, the whole thing would re render when we hovered over a row. This splits out the row into its own component that handles hover state internally.

## Screenshots

**BEFORE**
Screen Shot 2021-10-28 at 5 40 52 PM

**AFTER**
Screen Shot 2021-10-28 at 5 40 23 PM


apacosta (approved):
Great Improvement 💯

PR #246, opened 2021-12-20
Updated GridItem to use 24 instead of 12 columns

bolinkd:
## Description

Updated GridItem to use 24 instead of 12 columns

## Notes

Will be a major version change with a web PR.


jordanell (approved):
Obviously goes without saying, this is not backwards compatible. Will need to audit any packages for their use of GridItem when you update the dependency. You might want to actually do a minor version bump on riipen UI for this instead of the normal patch level.

ericauv (approved):

PR #250, opened 2022-01-20
[PLT-9020] Wrap AccordionDetails in AccordionContext

ericauv:
## Description
We want to enable AccordionDetails to be able to
access the Accordion context so that AccordionDetails
and its children can know whether or not their
Accordion is expanded. This can be used so that
we only load data when the accordion is expanded.

### DIFF
--- a/packages/riipen-ui/src/components/Accordion.jsx
+++ b/packages/riipen-ui/src/components/Accordion.jsx
@@ -14,7 +14,6 @@ const Accordion = props => {
     classes,
     defaultExpanded,
     disabled,
-    provideContextToChildren,
     ...other
   } = props;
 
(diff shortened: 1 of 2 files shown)


packages/riipen-ui/src/components/Accordion.jsx
```diff
@@ -48,17 +49,35 @@ const Accordion = props => {
...
-      </AccordionContext.Provider>
-      <Collapse in={expanded}>
-        <div
-          aria-labelledby={summary.props.id}
-          id={summary.props["aria-controls"]}
-        >
-          {children}
-        </div>
-      </Collapse>
+      {provideContextToChildren && (
```
bolinkd, line 52: im not super in on the Context stuff, what would be the downside of always wrapping the children with the context?

bolinkd:
Just confirming there was no noticeable side effect from these changes

bolinkd (approved):

ericauv:
@bolinkd  correct -- no noticeable changes from wrapping all accordion children in contexts (checked course examples, course deliverables, team batch)

PR #251, opened 2022-01-20
[PLT-8330] Change spacing from 5 to 4

apacosta:
## Description

Update spacing variable from 5 to 4


bolinkd (approved):
Be careful when merging this one, I know Eric is working on some ui kit stuff, so just make sure the versions line up correctly when deploying to make sure the right version makes it to production.

PR #252, opened 2022-01-25
NextJS 10

jordanell:
## Description

Downgraded nextjs to version 10 to properly support React 16 which is used in the repo. (NextJS 11 needs at least react 17 which was causing all sorts of issues).

Now following the instructions in the README works again.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,7 @@
 				"enzyme-to-json": "3.6.1",
 				"lodash": "4.17.21",
 				"marked": "^0.7.0",
-				"next": "11.1.1",
+				"next": "10.2.3",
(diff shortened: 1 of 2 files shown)


ksmort (approved):
Darn that sucks
How awful would it be to upgrade to React 17?

jordanell:
@ksmort aweful, because you can only run 1 version of react in any project, and major versions of react are not compatible with a ton of stuff, so you would need to upgrade all of Riipen's repos at once to do it.

ksmort:
> @ksmort aweful, because you can only run 1 version of react in any project, and major versions of react are not compatible with a ton of stuff, so you would need to upgrade all of Riipen's repos at once to do it.

😢

PR #255, opened 2022-01-26
Example for Kyubin

ericauv:
## Description
@Kyubinhan 

Example of adding a new component to docs


ericauv:
Is an example pr, not meant to be merged

PR #258, opened 2022-01-27
Remove python libs install steps in .circleci/config.yml

Kyubinhan:
## Description
Remove python libs install steps in the docs job cause they are not needed anymore.


stredger (approved):
probably dont need the update either, but for sure dont need the python anymore

PR #263, opened 2022-02-01
Allow tab to be prop type node

ksmort:
## Description
Update Tab to allow icons of type node.

## Where to Start
src/components/Tab

### DIFF
--- a/packages/riipen-ui/src/components/Tab.jsx
+++ b/packages/riipen-ui/src/components/Tab.jsx
@@ -75,7 +75,11 @@ const Tab = props => {
         onFocus={handleFocus}
         onBlur={handleBlur}
       >
-        {Icon && <Icon className={clsx("icon")} />}
+        {Icon && (
+          <span className={clsx("icon")}>
+            {typeof Icon === "function" ? <Icon /> : Icon}
+          </span>
+        )}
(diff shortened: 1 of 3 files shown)


stredger (approved):

packages/riipen-ui/src/components/Tab.jsx
```diff
@@ -75,7 +75,11 @@ const Tab = props => {
         onFocus={handleFocus}
         onBlur={handleBlur}
       >
-        {Icon && <Icon className={clsx("icon")} />}
+        {Icon && (
+          <span className={clsx("icon")}>
+            {typeof Icon === "function" ? <Icon /> : Icon}
+          </span>
+        )}
```
bolinkd, line 82: I would argue that we should change all implementations over to use
{ icon }
rather than the 
(props) => <Icon... />
stuff that we are currently using for the dashboard (I am assuming thats what we are doing with this change). As this way would be more "react-ful"

ksmort: Yep, that's the plan!  But I want the backwards compatibility of `function` until that has all been completed

bolinkd (approved):

PR #264, opened 2022-02-02
Improve tooltip arrow bottom position

apacosta:
## Description
Add border-width and calculated margin-top to Tooltip bottom arrow position to make it look better

## Notes
Is a small change but looks prettier

## Screenshots
Before:
Screen Shot 2022-02-02 at 12 03 24 PM
Screen Shot 2022-02-02 at 12 10 25 PM

After:
Screen Shot 2022-02-02 at 12 03 16 PM
Screen Shot 2022-02-02 at 12 09 50 PM


ksmort (approved):
🚀

PR #265, opened 2022-02-15
[PLT-9387] Add ButtonBase component

ericauv:
## Description
Replaces https://github.com/riipen/platform-web/pull/5045


packages/riipen-ui/src/components/AsButton.test.jsx
```diff
@@ -0,0 +1,133 @@
+import { mount } from "enzyme";
+import toJson from "enzyme-to-json";
+import React from "react";
+
+import AsButton from "./AsButton";
+
+describe("AsButton component", () => {
+  const child = <div>A child</div>;
```
ksmort, line 8:
```suggestion
let props;

beforeEach(() => {
props = { ariaLabel: ..., ..., children: <div>A child</div>, ...}
})
  const child = <div>A child</div>;
```

packages/riipen-ui/src/components/AsButton.test.jsx
```diff
@@ -0,0 +1,133 @@
...
+        wrapper
+          .find("AsButton")
+          .childAt(0)
+          .hasClass(classVariant[0])
+      ).toEqual(true);
+    });
+  });
+
+  describe("clearButtonStyle prop", () => {
+    it("Does not apply clearButtonStyle class when clearButtonStyle prop is false", () => {
```
ksmort, line 71: super duper nitpicky but should 'Does' be 'does' (lowercase)? Not sure what conventions are/if it matters capitalization/not for test description

packages/riipen-ui/src/components/AsButton.jsx
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * The children to render within the button
+   */
+  children: PropTypes.node.isRequired,
+
+  /**
+   * List of additional classes to apply to this component.
+   */
+  classes: PropTypes.array,
```
ksmort, line 61: array or string

packages/riipen-ui/src/components/AsButton.jsx
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * Whether traditional button styles (border, background) should be cleared
+   */
+  clearButtonStyle: PropTypes.bool,
```
ksmort, line 66: wonder if this should be `variant` with type `clear` or `default`?

packages/riipen-ui/src/components/AsButton.jsx
```diff
@@ -0,0 +1,82 @@
+import clsx from "clsx";
+import PropTypes from "prop-types";
+import React from "react";
+import _JSXStyle from "styled-jsx/style";
+
+const AsButton = ({
```
ksmort, line 6: I wonder if we could do something like https://mui.com/api/button-unstyled/ or https://mui.com/api/button-base/ in terms of name (asButton seems too much like a HOC).
Also if we're using for TemplateCards, is a button the best choice?  Seems like a lot to put into a <button> but tbh I'm not sure if there's rules or conventions so that's just an opinion with no backing evidence :P

ksmort (changes requested):
Couple nitpicks and a couple questions!

packages/riipen-ui/src/components/AsButton.jsx
bolinkd: I would agree with something like ButtonBase, seeing as we are really wrapping whatever we have in a button, rather than transforming the entity into a button

packages/riipen-ui/src/components/ButtonBase.jsx
```diff
@@ -0,0 +1,84 @@
...
+
+        .button {
+          cursor: pointer;
+        }
+
+        .button:focus {
+          outline: 5px auto -webkit-focus-ring-color;
+        }
+
+        .default {
```
ericauv, line 37: as this is the intended/most likely to be used variant, I made default be the variant with cleared button styles

packages/riipen-ui/src/components/ButtonBase.jsx
```diff
@@ -0,0 +1,84 @@
+import clsx from "clsx";
+import PropTypes from "prop-types";
+import React from "react";
+import _JSXStyle from "styled-jsx/style";
+
+const ButtonBase = ({
+  ariaLabel,
+  classes,
+  children,
```
ksmort, line 9: alphabetize

packages/riipen-ui/src/components/ButtonBase.jsx
```diff
@@ -0,0 +1,84 @@
...
+      <button
+        aria-label={ariaLabel}
+        className={className}
+        type={type}
+        {...other}
+      >
+        {children}
+      </button>
+      <style jsx>{`
+        .basic {
```
ksmort, line 26: I wonder if we can just remove basic css class since it's empty. What is the difference between default and basic?

packages/riipen-ui/src/components/ButtonBase.test.jsx
```diff
@@ -0,0 +1,110 @@
...
+  let props;
+  let children;
+
+  beforeEach(() => {
+    children = <div>A child</div>;
+    props = {
+      ariaLabel: "click to be cool"
+    };
+  });
+  it("renders without errors", () => {
```
ksmort, line 17:
```suggestion
  });
  
  it("renders without errors", () => {
```

ksmort (changes requested):
🚀 Just a couple nitpicks!

packages/riipen-ui/src/components/ButtonBase.jsx
ericauv: `basic` would make the wrapped children look like a basic html `button`
`default` removes the styles of standard html `button` element so that this component becomes an unstyled wrapper for its children

packages/riipen-ui-docs/src/pages/components-api/button-base.md
```diff
@@ -0,0 +1,30 @@
+<!--- This documentation is automatically generated, do not try to edit it. -->
+
+# AsButton API
+
+<p class="description">The API documentation of the AsButton React component.</p>
+
+## Import
+
+```js
+import AsButton from 'riipen-ui/AsButton';
```
bolinkd, line 10: this needs to be updated I believe

PR #269, opened 2022-03-01
[PLT-9035] more fixes to typography sizes on mobile

apacosta:
https://riipen.atlassian.net/browse/PLT-9035

## Description
- Add mobile font sizes for Badge, Input, MenuItem components and Tab default letter spacing to 1

### DIFF
--- a/packages/riipen-ui/src/components/Badge.jsx
+++ b/packages/riipen-ui/src/components/Badge.jsx
@@ -258,6 +258,17 @@ class Badge extends React.Component {
             transform: scale(1) translate(-50%, 50%);
             transform-origin: 0% 100%;
           }
+
+          @media (max-width: ${theme.breakpoints.sm}px) {
+            .badge.medium {
+              font-size: ${theme.typography.body2.mobile.fontSize};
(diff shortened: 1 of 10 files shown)


packages/riipen-ui/src/components/Typography.jsx
```diff
@@ -324,6 +331,13 @@ class Typography extends React.Component {
             font-weight: ${theme.typography.fontWeight.regular};
           }
 
+          .letter-spacing-1 {
+            letter-spacing: 1px;
```
ksmort, line 335: You could maybe take it one step further and do 
.letter-spacing {
 letter-spacing: ${letterSpacing}px;
}

And keep the boolean at line 170 but do
letterSpacing ? 'letter-spacing' : null

ksmort (approved):

packages/riipen-ui/src/components/Typography.jsx
apacosta: That sounds good thanks for the suggestion! 👏🏻

PR #276, opened 2022-04-07
[[PLT-9619] Remove fixed width size from RadioButton

apacosta:
## Description
Issue: fixed radio button width size make long label text jump a line.

- Remove fixed width size from RadioButton
- Fix RadioButton font medium size

### DIFF
--- a/packages/riipen-ui/src/components/RadioButton.jsx
+++ b/packages/riipen-ui/src/components/RadioButton.jsx
@@ -22,9 +22,11 @@ const RadioButton = ({
   const theme = React.useContext(ThemeContext);
 
   let typographyVariant;
-  if (size === "small") {
+  if (size === "medium") {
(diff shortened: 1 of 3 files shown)


packages/riipen-ui/src/components/RadioButton.jsx
```diff
@@ -22,9 +22,11 @@ const RadioButton = ({
   const theme = React.useContext(ThemeContext);
 
   let typographyVariant;
-  if (size === "small") {
+  if (size === "medium") {
+    typographyVariant = "body2";
+  } else if(size === "small"){
     typographyVariant = "body3";
-  } else if (size === "large") {
+  }else if (size === "large") {
```
ksmort, line 29:
```suggestion
  } else if (size === "large") {
```

ksmort (approved):

